PR #4, opened 2023-10-26
Reduce Dockerfile, require Node 18 or 20, prepare for monorepo release

eyelidlessness:
This PR contains the following:

- Semver major releases for all packages (because, at least, Node version bumps)
- Update main CI GitHub Actions config to restore publishing of NPM packages.
- Update and restore the Docker Hub and GHRC GitHub Actions.
- Remove everything concerning Vagrant, Docker Compose, and Heroku.
- Remove `pm2` and config templating from Enketo Express's Dockerfile.
- Update that Dockerfile to the latest supported Node version, and to use Yarn.
- Fixed a mistake in the monorepo transition, to ensure the extended date picker widget correctly includes styles from `bootstrap-datepicker`.
- Consolidate and clarify release instructions in the top-level README. I believe this accounts for the few special considerations between packages. It also includes some notes particular to the monorepo, mainly about steps where package order matters.

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

I've also verified the Dockerfile builds and runs. To do so locally, all of the following were necessary:

1. Run in an amd64 VM (arm64 did not work, Puppeteer fails to download Chrome)
2. Run said VM without Rosetta (Sass will build, but will crash when used)
3. Run Docker build with the platform explicitly specified as `linux/amd64` (otherwise it will still attempt to run as arm64)

This shouldn't matter for deployments, but worth noting for other developers.


packages/enketo-express/Dockerfile
```diff
@@ -1,18 +1,16 @@
...
 WORKDIR ${ENKETO_SRC_DIR}
 
-RUN npm install -g pm2@$(npm info pm2 version)
-
 COPY . ${ENKETO_SRC_DIR}
 
-RUN npm install -g npm@^6 && npm clean-install && npx grunt && npm prune --production
+RUN yarn install --frozen-lockfile && yarn grunt && yarn install --production
 
 # Persist the `secrets` directory so the encryption key remains consistent.
```
lognaturel, line 10: This is no longer needed -- we're not doing secrets management at all. Users are responsible for putting their own secrets in the config file.

lognaturel (approved):
I've verified that I can `yarn workspace enketo-express start` locally, run tests, and swapped in the image in a running ODK Central instance.

PR #7, opened 2023-10-30
Fix: check `github.ref` on release

eyelidlessness:
I tried this in a pared down action on a personal repo. For future reference:

- I saw it suggested in a Stack Overflow post that this would not work because the action isn't run on a `tags` event, but it did work in my testing
- There was some hint in various corners of documentation/intellisense/searching that one or several of the following might work (they all do not):
  - `github.event.inputs.ref`
  - `github.event.release.id`
  - `github.event.release.release_id`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
 
             - run: 'yarn workspace openrosa-xpath-evaluator test'
 
-            - if: github.event_name == 'release' && github.event.action == 'published' && startsWith(github.event.ref, 'refs/tags/openrosa-xpath-evaluator/') && matrix.node-version == '20.5.1'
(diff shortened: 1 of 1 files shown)


eyelidlessness:
On the latter two, [the docs](https://docs.github.com/en/rest/overview/github-event-types?apiVersion=2022-11-28#releaseevent) suggest there is a `github.event.release` object corresponding to the linked [release object](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-a-release). It's unclear what, if anything, exists on that object (nor whether it even exists, as my testing also revealed you can generally just dot-access into arbitrary non-objects without error). But also noting it here for posterity in case we ever do need to know what's there.

lognaturel (approved):

PR #1290, opened 2023-12-13
fixed: a loaded drawing/signature/annotation gets updated when nothin…

### LINKED ISSUE #1288: drawing changes when loaded from a record
**Describe the bug**

When loading a record that contains a drawing, the act of loading it changes the drawing file (name), even when the user makes no changes to it.

**To Reproduce**
Steps to reproduce the behavior. 

1. Load a record containing a drawing (widget) with a drawing file. Here is a [test form](https://docs.google.com/spreadsheets/d/1WlWhO8D2DAffkGT9EpnVY-lMu6Tmm_M1WDLwtP7kC-A/edit?usp=sharing)
2. Submit
3. Note how the drawing file was updated in the submission (e.g. by looking at the file name).

**Expected behavior**
For the drawing widget to leave a loaded drawing unchanged if the user makes no changes to it.

**Additional context**
~~I think the drawing widget revamp in March 2023, omitted the mechanism that set `this.value` in the drawing widget upon loading an existing drawing. This served to make sure no change event was propagated in `_updateValue` for loading an existing drawing. I don't immediately see the relevant code change causing this though.~~

MartijnR:
…g changed, closes #1288

-   [x] loading and manually updating drawings

see comment below

any form with a drawing/annotation/signature should work

### DIFF
--- a/packages/enketo-core/src/widget/draw/draw-widget.js
+++ b/packages/enketo-core/src/widget/draw/draw-widget.js
@@ -49,7 +49,7 @@ class DrawWidget extends Widget {
                 penColor: that.props.colors[0] || 'black',
             });
 
-            this.pad.addEventListener('change', () => {
+            this.pad.addEventListener('endStroke', () => {
                 this._updateValue();
             });
 


packages/enketo-core/src/widget/draw/draw-widget.js
```diff
@@ -49,7 +49,7 @@ class DrawWidget extends Widget {
                 penColor: that.props.colors[0] || 'black',
             });
 
-            this.pad.addEventListener('change', () => {
+            this.pad.addEventListener('endStroke', () => {
```
MartijnR, line 52: This was the change in https://github.com/enketo/enketo/commit/4f6abec01a9aae681df8dfb3ef0efc5b3dbda5e4# that caused the regression. It appears that the change event (instead of endStroke) works, but is not supported by signaturepad (https://github.com/szimek/signature_pad?tab=readme-ov-file#events). The problem with using it is that `this.pad.off()` doesn't actually turn it off. I don't know if there was a specific reason for using 'change' though .

lognaturel: Ah, no, I looked a little deeper and that doesn't make sense either.

lognaturel: Can we just move the event listener below the `if (existingFilename)` block? The problem is that setting the existing image registers as a change, right?

MartijnR: Oops! Sorry about that attempt to revert that key fix/feature!  I also looked deeper and also concluded extending `off()` wouldn't work. We could turn it off in draw-widget.js with removeEventListener(). Might have to be done at several places (and turned on again in several places). Or in the `ResizableSignaturePad` by not firing a change event if it's `off` (but not sure if we can easily create an `isOff()` check).

MartijnR: I'll revisit this. Probably not until next week though. Thanks for reviewing!

eyelidlessness:
I took a look into addressing this, with mixed success. My changes are [here](https://github.com/enketo/enketo/compare/fix/draw-widget-edit-submitted-image-no-update-on-load). The general idea is to distinguish between:

1. Loading image data, and loading it as the `ResizableSignaturePad` base image
2. Loading it as the base image on init, and subsequently (a proper change warranting dispatch of the `change` event[^1])

The changes address #1288, but causes data loss when editing submissions in a paginated form (where the draw widget is loaded on anything but the last page). I'm reasonably certain this is related to the outstanding bug alluded to in [this comment](https://github.com/enketo/enketo/blob/87d87ca582b56207854fedacaf81359b1475e1d0/packages/enketo-core/src/widget/draw/ResizableSignaturePad.js#L408-L428).

[^1]: Since there was some confusion about that, it bears clarification that the `change` event is dispatched by `ResizableSignaturePad` itself, not by the underlying signature-pad library/`SignaturePad` parent class.

MartijnR:
Thanks @eyelidlessness! I like your approach. Wrt the remaining issue, I think something is not going as planned in the file-manager. It wasn't supposed to extract the canvas images of unchanged canvases, I think. That's why [this] is there, I think,  to add those filenames to the files array. Still looking...

MartijnR:
> It wasn't supposed to extract the canvas images of unchanged canvases, I think.

It seems like it was, but those unchanged drawings would then be stripped from the submission because the form model [never set the file attribute on that XML node](https://github.com/enketo/enketo/blob/main/packages/enketo-core/src/js/nodeset.js#L148) because the value never changed. I don't remember why it was done in this manner, and it seems clear things are messy (artefacts or messily designed in the first place by somebody who will remain unnamed... ;)). That mechanism seems to still be in place as far as I can see.

So when looking at actual submissions, I can actually not reproduce the remaining bug you mentioned in your footnote. Hopefully that is good news, but likely you have a different way of revealing it after all...

lognaturel:
> So when looking at actual submissions, I can actually not reproduce the remaining bug you mentioned in your footnote.

Do you mean the problem with the background/previously saved image not being part of the submitted image? If so, are you using the form linked above?

Or is this maybe about some other specific aspect of the possible file-manager issues?

MartijnR:
> Do you mean the problem with the background/previously saved image not being part of the submitted image? 

Oh, is that is what is meant with data loss (I thought it had to do with image resolution or filename changes). In that case I can reproduce it but I thought it was a (longstanding) feature to not submit unchanged attachments instead of a bug. Hmm.

lognaturel:
There are a lot of different pieces! The comment in the code on master has to do with resizing. But that code also is too aggressive about detecting changes and considers setting an existing image as a change. That’s the bug we’re trying to fix here. 

@eyelidlessness’ fix above looks like it works perfectly EXCEPT for the case where there are images on pages other than the last one and those images have a default background or an existing value loaded in. In those cases, the background or prior data is not included in a resubmission. That’s far worse data loss.

lognaturel:
> feature to not submit unchanged attachments

Yes, agreed that should be the case and is not on master currently. (To be explicit!)

MartijnR:
>  In those cases, the background or prior data is not included in a resubmission. That’s far worse data loss.

Thanks for explaining. Yes, that's a serious one indeed.

lognaturel:
We've decided to revert the change to preserve annotation resolution for now https://github.com/enketo/enketo/pull/1292

PR #1293, opened 2024-01-26
Prepare releases for enketo-core and enketo-express

lognaturel:
I did not do a full dependency update because it resulted in some lint and test failures I'm not best positioned to dive into.

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Run on a server as part of the Central stack and verified that basic functionality works. Also verified that annotation implementation is reverted as we expect.

I could have spent more time trying to update dependencies but most of the ones that are out of date are dev dependencies without security issues so making sure `vite` is upgraded seems sufficient.

This is a small release primarily to revert the annotation implementation. It should be low risk.

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,10 +92,10 @@
         "sinon-chai": "^3.7.0",
         "time-grunt": "^2.0.0",
         "typescript": "^5.2.2",
-        "vite": "^4.4.9",
-        "vite-node": "^0.34.3",
+        "vite": "^4.5.2",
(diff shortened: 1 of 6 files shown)


yanokwa (approved):
My review of vulns on Snyk don't list vite, so I don't know how necessary that change is. @lognaturel says it's not a risky upgrade because it's a build component, so I'm happy to proceed.

PR #1296, opened 2024-01-29
Fixed _isRelevant function in calculation module returns incorrect ca…

MartijnR:
…ched result after initialization

It's a bit of an obscure issue due to the fact that in most of the Enketo world form values are never cleared when they become non-relevant. However, there are some remnants in the code (e.g. the `emptyNonRelevant` parameters at various places in calculate.js), that conflict with the `preInitRelevance` cache in calculate.js. See inline code comment below.

I had to write a unit test to demonstrate it and decided to do this at the lowest level since `_isRelevant` should probably always return a correct result no matter when it is called (so the test would survive any possible future removal of these emptyNonRelevant remnants). Though if you prefer I write a test for the `update` method, just let me know.

### DIFF
--- a/packages/enketo-core/src/js/calculate.js
+++ b/packages/enketo-core/src/js/calculate.js
@@ -482,7 +482,7 @@ export default {
                         const preInitRelevance =
                             this.preInitRelevance.get(element);
 
-                        if (preInitRelevance != null) {
+                        if (preInitRelevance != null && !this.initialized ) {
                             return preInitRelevance;
                         }
 
--- a/packages/enketo-core/src/js/calculate.js
+++ b/packages/enketo-core/src/js/calculate.js
@@ -482,7 +482,7 @@ export default {
                         const preInitRelevance =
                             this.preInitRelevance.get(element);
 
-                        if (preInitRelevance != null && !this.initialized ) {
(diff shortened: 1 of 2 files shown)


packages/enketo-core/src/js/calculate.js
```diff
@@ -482,7 +482,7 @@ export default {
                         const preInitRelevance =
                             this.preInitRelevance.get(element);
 
-                        if (preInitRelevance != null) {
+                        if (preInitRelevance != null && !this.initialized ) {
```
MartijnR, line 485: I believe the intent of the preInitRelevance cache is to be used only during initialization and not afterwards.

lognaturel (approved):
Sorry about the delay on this, @MartijnR, it took some effort to wrap my head around. I am feeling confident that I understand it and that it is safe. The test looks like it would adequately catch regressions to me.

lognaturel:
One thing we haven't completed after the monorepo migration is hooking up benchmarks. We should do a benchmark sanity check before release.

MartijnR:
Thank you! 

> One thing we haven't completed after the monorepo migration is hooking up benchmarks. We should do a benchmark sanity check before release.

Ah, yes!

PR #1298, opened 2024-02-15
Add support to Redis db index

noliveleger:

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Monitor Redis to ensure it does send "select `<db index>`" before sending the command

Maybe some of my PR could be merged with #1297 since the latter gives more leeway when using redis URLs. 

There should not be any regression risk because it falls back on db index 0 like it is in the current release. However, it lets the sysadmin to use 2 different databases when using only one instance of redis.


lognaturel:
We'll use full URLs for any advanced configuration. Closing in favor of #1297

Issue #994, opened 2022-07-19
Cookies aren't forwarded when making HEAD request to get max-size from form server


KipSigei:
The `https://<server-url>/submission/max-size/<survey id>` endpoint seems to be failing with a 401 `Forbidden. Authorization Required` error.  The root cause for this is when enketo makes a `HEAD` request to the data server `enketo/<form-id>/submission` endpoint. Instead it shows the default `5mb` value.

![image](https://user-images.githubusercontent.com/19689992/179716673-25dfad04-8ee8-47a9-bfe7-12b27e06d2a9.png)

PR #1303, opened 2024-03-22
Forward auth cookie when making max-size request to form server

FrankApiyo:
Closes: https://github.com/enketo/enketo/issues/994
Closes: https://github.com/onaio/onadata/issues/2568

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Tested locally

Yes, I considered [allowing HEAD requests on this endpoint to be unauthenticated on onadata side
](https://github.com/onaio/onadata/pull/2565)
No regression risks that I can think of

No

### DIFF
--- a/packages/enketo-express/app/controllers/submission-controller.js
+++ b/packages/enketo-express/app/controllers/submission-controller.js
@@ -154,6 +154,7 @@ function maxSize(req, res, next) {
         surveyModel
             .get(req.enketoId)
             .then((survey) => {
+                survey.cookie = req.headers.cookie;
                 survey.credentials = userModel.getCredentials(req);
 
                 return survey;


FrankApiyo:
I noted while testing this locally that the HEAD request to the submission endpoint is sent without any authentication (set to `null`) and without any cookies. I believe this isn't the desired behaviour
```
{
  url: 'http://192.168.100.65:8000/enketo/9/submission',
  auth: null,
  headers: { cookie: undefined },
  method: 'head'
}
```

The fix presented here is to relay cookies sent from the browser to the ODK server

lognaturel:
You have a form that requires cookie-based auth and it looks like cookies aren't forwarded when making the HEAD request to get max-size from Enketo backend to the form server, is that right?

Do you believe this to be a change in behavior in Enketo? Is there some change on your end that revealed this now?

FrankApiyo:
This is right
> You have a form that requires cookie-based auth and it looks like cookies aren't forwarded when making the HEAD request to get max-size from Enketo backend to the form server, is that right?

No, this is not a change in behavior in enketo. It seems that the max-size endpoint has always worked like this (the request has always been made without credentials).

lognaturel:
Great, thanks for the background. It looks like it was probably an accidental oversight but I do want to give it some thought and look at the code around this area more carefully since this is auth-related.

lognaturel (approved):
This excludes the direct XForm URL case and matches the behavior with basic/digest auth.

PR #1304, opened 2024-04-16
Update dependencies

lognaturel:
In preparation for release later this week.

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

I decided to switch over to @cypress/request to address security issues. Cypress' fork is mentioned in the Dependabot report for those issues.

There's a bug in Firefox that made one of the Transformer tests fail (https://github.com/enketo/enketo/pull/1304#issuecomment-2059700645). I decided to add a similar case and comment out the failing one for now.

I tried to upgrade `vite` to 5.x but couldn't get past an import issue just in the browser tests. I think it might be https://github.com/vitejs/vite/issues/6582? I initially thought it could be related to the rollup upgrade (https://vitejs.dev/guide/migration#rollup-4) but I don't think so. Someone who knows more about this tooling can likely figure it out quickly.


lognaturel:
There's been some kind of change in behavior in node between 20.5.1 and 20.12.2 that results in string values that end in a number preceded by a non-alphanumeric character passed to `date()` to convert to `true` in the invalid dates test.

`date(opv_3)` is now failing. So do `date(opv 3)` or `date(opv.3)` but not `date(opv3)`.

I see that the first three are being evaluated to `2001-03-01`. `date(afdf_4)` evaluates to `2001-04-01`. `date(jkfdk_32)` evaluates to `2032-01-01`.

I can see that in `asDate` from `openrosa-extensions` all of these values get passed to `new Date()`.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:20
+FROM node:20.12
```
alxndrsn, line 1: CI is very specific about version numbers.  Would it make sense for the version here to match the version tested against?

alxndrsn (approved):

Dockerfile
lognaturel: Our current approach for Docker images is to fix the minor version and let them get patches. That means in case of significant security issues in the image a rebuild is sufficient and we don’t have to rerelease. 

The downside is that the build isn’t entirely reproducible and, like you say, doesn’t exactly match CI which needs an explicit patch version. The alternative would be to use SHA hashes for images. 

I think where we’ve landed is a good balance. Maybe we go with this for now to match other parts of the system and we can have a broader discussion if you think we should go back to tighter reproducibility.

PR #1306, opened 2024-04-16
transformer/markdown: url-encode href attributes

alxndrsn:


packages/enketo-transformer/test/markdown.spec.ts
```diff
@@ -68,6 +68,18 @@ describe('markdown', () => {
                 '[link](http://example.org)',
                 '<a href="http://example.org" rel="noopener" target="_blank">link</a>',
             ],
+            [
+                '[link](http://example.org/cgi-bin/Markdown.pl?a=1&b=2)',
+                '<a href="http://example.org/cgi-bin/Markdown.pl?a=1&b=2" rel="noopener" target="_blank">link</a>',
+            ],
+            [
+                '[link with a "quote" inside](http://example.org)',
```
lognaturel, line 76: Did this case intend to have a quote inside?

alxndrsn: only in the link text... would `link text with "quotes" inside` be clearer?

lognaturel: The intent is to verify that the label is not uri-encoded?

lognaturel (approved):
I think there's some possibility that forms exist with URLs that have manually URL-encoded parts and those will now be double-encoded and stop working. That's unfortunate but this should have been the initial behavior so it feels right to change.

PR #1307, opened 2024-04-17
enketo-express/test/survey-model: reduce test flake

alxndrsn:
Occasional test failures can be seen in Survey Model tests, e.g. https://github.com/enketo/enketo/actions/runs/8701494028/job/23863630099:

```
1) Survey Model
       getList
         obtains the list surveys if all are active in ascending launch date order:

      AssertionError: expected [ { …(2) }, { …(2) }, { …(2) } ] to deeply equal [ { …(2) }, { …(2) }, { …(2) } ]
      + expected - actual

           "openRosaId": "a"
           "openRosaServer": "https://kobotoolbox.org/enketo"
         }
         {
      +    "openRosaId": "b"
      +    "openRosaServer": "https://kobotoolbox.org/enketo"
      +  }
      +  {
           "openRosaId": "c"
           "openRosaServer": "https://kobotoolbox.org/enketo/deep"
         }
      -  {
      -    "openRosaId": "b"
      -    "openRosaServer": "https://kobotoolbox.org/enketo"
      -  }
       ]

      at /home/runner/work/enketo/enketo/node_modules/chai-as-promised/lib/chai-as-promised.js:302:22
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

This appears to be due to timestamp collisions causing `surveys.sort(_ascendingLaunchDate)` to be unstable.

Creating two Date objects 1ms apart in javascript does not guarantee they will have been created in different clock milliseconds.  This can be demonstrated:

```js
const DELAY = 1;
let start, end, iterations;
iterations = 0;
const recordEnd = () => {
  end = new Date();
  if(start.toISOString() === end.toISOString()) {
    throw new Error(`Start and end matched after ${iterations} iterations`);
  } else repeat();
};
const repeat = () => {
  ++iterations;
  start = new Date();
  setTimeout(recordEnd, DELAY);
};
repeat();
```

On my machine, a 1ms delay will cause failure after ~10 iterations.  A 2ms delay will not fail before I get bored of waiting.

### DIFF
--- a/packages/enketo-express/test/server/survey-model.spec.js
+++ b/packages/enketo-express/test/server/survey-model.spec.js
@@ -12,11 +12,12 @@ const { expect } = chai;
 
 // help function to ensure subsequent database entries don't have the exact same timestamp
 // redis is fast...
-const _wait1ms = () =>
+// 1ms is not enough to guarantee ISO timestamps are different.
+const _wait2ms = () =>
     new Promise((resolve) => {
         setTimeout(() => {
             resolve();
-        }, 1);
+        }, 2);
     });
 
 describe('Survey Model', () => {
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Let's see if "Alex gets bored of waiting" is long enough for this use case... 😄

PR #1309, opened 2024-04-18
Use slim node image, clean cache

lognaturel:
Reduce the image size. It was over 1.6Gb compressed because of the double `yarn install`. The second `yarn install` was not doing anything: `yarn` v1 doesn't do any pruning of existing installed deps with `--production`: https://github.com/yarnpkg/yarn/issues/6373, https://github.com/yarnpkg/yarn/issues/696

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Tried integrated with ODK Central.

There's more to do but I think we should start with this which gets to a compressed image of about 800Mb. This makes a big difference in install time, even on fast connections.

I had hoped to prune dev dependencies and I think https://github.com/enketo/enketo/pull/1309/commits/2469791fdb15ecba8ff839c180e6ee9f5da6baba is a reasonable way but @alxndrsn is less comfortable with it. I'd rather go with an approach we all feel is safe, the diff is 70mb and the dev dependencies should not pose any risk. I think the next step will be to come back to this and take a multistage approach.

I think given that I've run it with a form server the risks are very low. The changes are all implementation details and shouldn't change anything for downstream users.

No.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM node:20.12.2-slim
 
 RUN apt-get update || : && apt-get install -y \
     python3 \
-    build-essential
+    build-essential \
+    jq
 
 ENV ENKETO_SRC_DIR=/srv/src/enketo
 WORKDIR ${ENKETO_SRC_DIR}
@@ -11,7 +12,7 @@ COPY . ${ENKETO_SRC_DIR}
 
 # Install and build, first with dev dependencies, then with production only (yarn 1 has no prune)
 RUN yarn install --frozen-lockfile \
-    && rm -rf node_modules \
+    && yarn remove $(cat package.json | jq -r '.devDependencies | keys | join(" ")') -W \
     && yarn install --production --frozen-lockfile --ignore-scripts \
     && yarn cache clean
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,10 +10,8 @@ WORKDIR ${ENKETO_SRC_DIR}
 
 COPY . ${ENKETO_SRC_DIR}
 
-# Install and build, first with dev dependencies, then with production only (yarn 1 has no prune)
+# Install and build, leaving dev dependencies (yarn 1 has no prune)
 RUN yarn install --frozen-lockfile \
-    && yarn remove $(cat package.json | jq -r '.devDependencies | keys | join(" ")') -W \
-    && yarn install --production --frozen-lockfile --ignore-scripts \
     && yarn cache clean
 
 EXPOSE 8005
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM node:20.12.2-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         python3 \
-        build-essential
+        build-essential \
+        git
 
 ENV ENKETO_SRC_DIR=/srv/src/enketo
 WORKDIR ${ENKETO_SRC_DIR}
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,15 +1,20 @@
-FROM node:20.12.2
+FROM node:20.12.2-slim
+
+RUN apt-get update || : && apt-get install -y \
```
alxndrsn, line 3: `A || : B` seems obscure; is it possible just to write `A; B`?

alxndrsn: Also I'm not sure I've previously seen `apt-get update` failures handled in Dockerfiles before; is there a reason specifically to use it here?

Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 COPY . ${ENKETO_SRC_DIR}
 
-# Install (including dev dependencies) and build
-RUN yarn install --frozen-lockfile
-
-# Remove dev dependencies after build
-RUN yarn install --production --frozen-lockfile --ignore-scripts
+# Install and build, first with dev dependencies, then with production only (yarn 1 has no prune)
+RUN yarn install --frozen-lockfile \
+    && yarn remove $(cat package.json | jq -r '.devDependencies | keys | join(" ")') -W \
```
alxndrsn, line 15: This step wasn't present previously; has something changed?  It also looks like it might modify `yarn.lock`, and so potentially cause the `yarn install --production` to resolve dependencies differently to those expected...

Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 
 COPY . ${ENKETO_SRC_DIR}
 
-# Install (including dev dependencies) and build
-RUN yarn install --frozen-lockfile
-
-# Remove dev dependencies after build
-RUN yarn install --production --frozen-lockfile --ignore-scripts
+# Install and build, first with dev dependencies, then with production only (yarn 1 has no prune)
+RUN yarn install --frozen-lockfile \
```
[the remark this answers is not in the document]
lognaturel: This is weird, I agree. I decided to leave it because I don’t want to build yet another time. All of the necessary packages end up being built because of the dependencies between them.

Dockerfile
lognaturel: I copied this from an example on switching to a slim image. But can simplify!

Dockerfile
lognaturel: It doesn’t work for reasons I don’t entirely understand. It looks like the native dependencies don’t get rebuilt for some reason.

alxndrsn: because `--ignore-scripts`?

lognaturel: Maybe so. But the other thing is that I’d rather we not have to rebuild them. It’s time consuming and feels generally silly to do that work multiple times.

lognaturel: @alxndrsn directly shared with me a concern that `yarn remove` might rebuild the whole lock file. I've confirmed that `yarn remove` surgically removes the dependency(ies) specified.

lognaturel: I think the `yarn remove` step is safe but it certainly is not the most elegant thing. Let's remove for now and keep dev dependencies. It still halves the image size and the dev dependencies were included before anyway.

yanokwa (approved):
I ran https://github.com/wagoodman/dive on this and it says it's very efficient (as far as Docker is concerned). 

You could add `&& apt-get autoremove && apt-get clean` to the first RUN to clean up everything, but https://docs.docker.com/develop/develop-images/instructions/ says that images like node slim based on Debian already do that.

I had to add `ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true` to get the thing to build on my M1 Mac.

PR #1327, opened 2024-07-11
[Form redirection option] - Reopen empty form after submit

PNPyrenees:
Closes #

-   [ ] Saving offline drafts
-   [x] Editing submissions - But keep original redirection to the thanks page because in editing redirection already made (I don't know really where...)
-   [x] Form preview
-   [ ] None of the above

Admin can choose if he want to active or not the redirection to a new empty form by specifying the value of "redirectToForm" in enketo-express/config/config.json.template file

If redirectToForm = true, the user can submit more easly many forms. He don't need to click again on public url to go back on the form page and submit another.
.

### DIFF
--- a/packages/enketo-express/app/models/config-model.js
+++ b/packages/enketo-express/app/models/config-model.js
@@ -349,6 +349,7 @@ module.exports = {
         csrfCookieName: config['csrf cookie name'],
         excludeNonRelevant: config['exclude non-relevant'],
         experimentalOptimizations: config['experimental optimizations'],
+        redirectToForm: config.redirectToForm,
     },
     getThemesSupported,
 };
--- a/packages/enketo-express/config/default-config.json
+++ b/packages/enketo-express/config/default-config.json
@@ -123,5 +123,6 @@
     "exclude non-relevant": false,
     "experimental optimizations": {
         "computeAsync": false
-    }
+    },
+    "redirectToForm": false
 }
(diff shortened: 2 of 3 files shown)


MartijnR:
Just checking if you saw this: https://github.com/enketo/enketo/blob/main/packages/enketo-express/public/js/src/module/controller-webform.js#L346-L349

Some webform views in Enketo don't have redirects. The ones that meant to be used to submit multiple records in a row do not. Only the ones that are meant for _single submissions_ have a redirect (and I think that still makes sense or doesn't it?)

Would it perhaps be better to just use a different view (different API endpoint)? You may have seen https://apidocs.enketo.org/v2. I suspect you are using one of the `../single/..` endpoints. Using e.g. `POST /survey` or `POST /survey/offline` might provide the webform view that you are looking for.

lognaturel:
Bonjour @PNPyrenees!

Some of the language you use makes me think you may be using an ODK Central Public Access Link. Public Access Links use what Enketo calls `single` links as @MartijnR describes because these are typically intended for self-report. If you don't want that behavior, you can manually remove the `/single/` part of the URL before distributing your link as long as the link is configured without "single submission."

Hopefully these ideas help you make progress. If you want to discuss your needs in greater detail, I encourage you to start a thread at https://forum.getodk.org/. It would be useful to understand how you're getting your Enketo links and how they're being used (e.g. by staff, by volunteers, by the public).

PNPyrenees:
Thank you a lot,
As you tell me, the solution was just to remove the "/single" in the public url given by ODK ! Wonderful !

This PR can be dropped.

This allowed me to know a little more about how Enketo works ;)

lognaturel:
Wonderful, I’m glad you got to a good solution! We’d love to see you on the forum (https://forum.getodk.org) if ever you have more questions or want to share what you’re using Enketo and ODK for. There are a number of users there working on France and specifically on conservation. Hope to see you there soon! 😉

Issue #997, opened 2022-07-05
Annotated image is resized according to browser display


lognaturel:
https://forum.getodk.org/t/difference-b-n-collect-enketo-low-image-resolution/38096

PR #1335, opened 2024-07-23
Simplify draw widget logic by not scaling image when canvas size changes

### LINKED ISSUE #1323: Draw image filename postfix changes on `canvasreload` event if image value did not change
**Describe the bug**
First off, I just want to confirm that this issue presents in the new (aka "old") version of the Draw widget for enketo-core `8.1.0+`.

Basically, the `_updateValue` function in the Draw widget will [always assign a new value](https://github.com/enketo/enketo/blob/05e5281949e54c0be6592996513ab592a7b55403/packages/enketo-core/src/widget/draw/draw-widget.js#L448) for the `filenamePostfix` every time the function is called with a different value in the pad.  However, the `originalInputValue` field is only reset if `changed === true` which is NOT the case if `_updateValue`  was called by [a `canvasreload` event](https://github.com/enketo/enketo/blob/05e5281949e54c0be6592996513ab592a7b55403/packages/enketo-core/src/widget/draw/draw-widget.js#L236).

Changing the `originalInputValue` is what causes the text filename _with the postfix_ to be saved in the form model as the "answer" for that question. However, when the postfix for the file changes and  `originalInputValue` is not updated (as in the case of the `canvasreload` event, then a mismatch occurs between the filename (e.g. as [returned by `fileManager.getCurrentFiles`](https://github.com/enketo/enketo/blob/05e5281949e54c0be6592996513ab592a7b55403/packages/enketo-core/src/js/file-manager.js#L133)) and the value recorded for that question in the model string (e.g. as returned by [`form.getDataStr()`](https://github.com/enketo/enketo/blob/05e5281949e54c0be6592996513ab592a7b55403/packages/enketo-core/src/js/form.js#L556)).  

**To Reproduce**
This behavior is not really noticeable by the user when filling out the form. The form can be filled out successfully. The problem occurs after submitting the form if anyone tries to correlate the string value saved for the Draw question (the filename+postfix) with the name of the actual image file recorded by the form. These will not match if the `canvasreload` event is triggered after the last drawing input is made on the canvas.  

Basically, to trigger the problem behavior:

1. Open a form with a draw widget
2. Draw on the canvas
3. Resize your webpage so that the size of the canvas changes
4. Complete the form and see that the name of the image file does not match name stored in the form model

Even though this behavior is not visible to the user, it is possible to see the problem behavior in action out on https://staging.enketo.getodk.org/.  

- I uploaded a form with a draw widget to https://getodk.org/xlsform/
- Then, after opening the form, I use my browser DevTools to place a breakpoint [in the `nodeset.setVal` function](https://github.com/enketo/enketo/blob/05e5281949e54c0be6592996513ab592a7b55403/packages/enketo-core/src/js/nodeset.js#L114)
- When I draw on the canvas, the `nodeset.setVal` function is triggered. I can see that `newVals` is always the filename + a new postfix value each time I draw on the the canvas.  
- However, when I change the size of my browser window so that the size of the canvas is updated, the `nodeset.setVal` function is not triggered.  BUT, looking at the actual HTML page body, I can see the `data-filename-postfix` value for the `input` of my draw question get updated as the canvas refreshes.

**Expected behavior**
I expect that the name of the image file attachment should match the string value recorded for the draw question in the model.  

Looking at the code for the `draw._updateValue` function, it seems like we could avoid resetting the `filenamePostfix` except when `changed === true` (and `originalInputValue` is getting reset).  I am not sure why we would need/want a different postfix for `canvasreload` events.  (But maybe I am missing something here???)  I have tried moving the postfix code inside the `if (changed)` block and it resolved the issue (and did not create any new issues that I could detect).  If this seems like a reasonable fix, I am happy to submit a PR!

- OS: Debian
- Browser: Chromium 124.0.6367.60 
- Enketo: Reproducible on https://staging.enketo.getodk.org/

jkuester:
Closes #1323

As @lognaturel [suggested](https://github.com/enketo/enketo/issues/1323#issuecomment-2195728964), I decided to take some time and try to address #1323 more holistically.  My main goal with these code changes was to find a workable middle ground between the current functionality and the more forward-thinking approach in https://github.com/enketo/enketo-core/pull/960.  Instead of re-working everything from the ground up, I instead tried to simplify the code flow as necessary while changing as little code as possible.  I have got the implementation code working just like I want it, (and the existing automated tests pass for me locally), but I have not yet added any additional automated tests. Before going any further, I am hoping to get some feedback on my approach.  

The main behavior change I am proposing here is to suggest that we avoid trying to _scale_ the image drawn on the canvas at all.  If the canvas size changes (e.g. when the window is resized), the available canvas that can be drawn on grows, but the existing drawing/image does not scale.  This is the approach taken by the [signature_pad example app](https://szimek.github.io/signature_pad/) and after consideration of various [documentation](https://github.com/szimek/signature_pad?tab=readme-ov-file#handling-canvas-resize), [code](https://github.com/szimek/signature_pad/blob/4aebbc60cc1efdbf7214bde379db7022da0311c0/js/app.js#L18), and [issues](https://github.com/szimek/signature_pad/issues/265) I have come to think it offers the best balance of functionality and simplicity of code.  Practically speaking this means that the size of what is drawn will remain the same regardless of how the size of the canvas changes. If the canvas becomes smaller, parts of the drawing/signature may no longer be visible, but they will still be included in the image that gets output upon form completion.  For annotations, the attached image will be scaled/centered based on the size of the canvas _when the image is selected_.  Subsequent changes to the canvas size will not cause the image to be re-scaled, but instead it will retain its proportional size just like the drawing.  

To be clear, I realize this is not a perfect solution for https://github.com/enketo/enketo/issues/997 since the attached image can still loose resolution if it is larger than the canvas is when it is attached. But, just protecting from extra resolution loss caused by changes to the canvas size seems like a step in the right direction....

[Screencast from 2024-07-23 17-02-31.webm](https://github.com/user-attachments/assets/708a4313-1fd9-41d4-b60e-21a36c0df894)

This PR:

- Updates the draw widget code to use the latest version of `signature_pad`.
- Removes the "custom" functions added to the `SignaturePad`. The updated logic is able to achieve the necessary functionality via just the actual SignaturePad api.
- Adds a `ResizeObserver` to handle triggering the canvas resize logic more efficiently  (instead of just listening to `window.on('resize'...`
- Removes `_updateWithDelay`/`_forceUpdate` logic.  It did not seem to be necessary now that the image value does not change then the canvas is resized.
- Added new context object `baseImage`.  This data is used to re-draw the annotation image when the canvas size changes:
  - `objectUrl` - the original image value attached to an annotation
  - `options` - the options values needed for drawing the image to the SignaturePad

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

I have, of course, extensively tested this functionality in the cht-core codebase (covering all the workflows from above that the CHT supports...).  I have also tested via the enekto-core development server with these forms:

[draw-forms.zip](https://github.com/user-attachments/files/16354675/draw-forms.zip)

Hopefully, my explanation above is sufficient.  

This does change the way the drawing widget works with regards to scaling the image when the window size changes.  However, in my opinion this is acceptable given that the current behavior is very poor (by making the canvas too small, you can significantly distort the image/drawing, even if you return the canvas to its original size).  

There is also a non-zero chance that this will somehow break other drawing related logic flow edge-cases. The existing widget flow is pretty sketchy and there have been a lot of "fixes" bolted on top over the years.  I tried to change as little as necessary here to hopefully minimize the risk of breaking random things.

Attached a group of forms above that I used for testing.

### DIFF
--- a/packages/enketo-core/package.json
+++ b/packages/enketo-core/package.json
@@ -62,7 +62,7 @@
         "mergexml": "1.2.3",
         "node-forge": "^1.3.1",
         "openrosa-xpath-evaluator": "3.1.0",
-        "signature_pad": "2.3.x"
+        "signature_pad": "^5.0.2"
     },
     "engines": {
         "node": ">=18 <21",
--- a/packages/enketo-core/src/js/file-manager.js
+++ b/packages/enketo-core/src/js/file-manager.js
@@ -13,7 +13,6 @@ import $ from 'jquery';
 
 import { t } from 'enketo/translator';
 import { getFilename, dataUriToBlobSync } from './utils';
-import DrawWidget from '../widget/draw/draw-widget';
 
 const fileManager = {};
 const URL_RE = /[a-zA-Z0-9+-.]+?:\/\//;
@@ -142,13 +141,12 @@ fileManager.getCurrentFiles = () => {
         .each(function () {
             let newFilename;
             let file = null;
-            let canvas = null;
             if (this.type === 'file') {
                 file = this.files[0]; // Why doesn't this fail for empty file inputs?
             } else if (
-                this.value
-                && !URL_RE.test(this.value)
-                && this.dataset?.cache
+                this.value &&
+                !URL_RE.test(this.value) &&
+                this.dataset?.cache
             ) {
                 // Load drawing from cache
                 file = dataUriToBlobSync(this.dataset.cache);
--- a/packages/enketo-core/src/js/file-manager.js
+++ b/packages/enketo-core/src/js/file-manager.js
@@ -13,6 +13,7 @@ import $ from 'jquery';
 
 import { t } from 'enketo/translator';
 import { getFilename, dataUriToBlobSync } from './utils';
+import DrawWidget from '../widget/draw/draw-widget';
 
 const fileManager = {};
 const URL_RE = /[a-zA-Z0-9+-.]+?:\/\//;
@@ -144,15 +145,14 @@ fileManager.getCurrentFiles = () => {
             let canvas = null;
             if (this.type === 'file') {
                 file = this.files[0]; // Why doesn't this fail for empty file inputs?
-            } else if (this.value) {
-                canvas = $(this)
-                    .closest('.question')[0]
-                    .querySelector('.draw-widget canvas');
-                if (canvas && !URL_RE.test(this.value)) {
-                    // TODO: In the future, we could simply do canvas.toBlob() instead
-                    file = dataUriToBlobSync(canvas.toDataURL());
-                    file.name = this.value;
-                }
+            } else if (
+                this.value
+                && !URL_RE.test(this.value)
+                && this.dataset?.cache
+            ) {
+                // Load drawing from cache
+                file = dataUriToBlobSync(this.dataset.cache);
+                file.name = this.value;
             }
             if (file && file.name) {
                 // Correct file names by adding a unique-ish postfix
--- a/packages/enketo-core/src/widget/draw/draw-widget.js
+++ b/packages/enketo-core/src/widget/draw/draw-widget.js
@@ -43,6 +43,7 @@ class DrawWidget extends Widget {
             '.draw-widget__body__canvas'
         );
         this.resizeObserver = new ResizeObserver(this._resizeCanvas.bind(this));
+        this.promisedRedrawPad = Promise.resolve();
 
         if (this.props.load) {
             this._handleFiles(existingFilename);
@@ -96,9 +97,9 @@ class DrawWidget extends Widget {
                         const data = that.pad.toData();
                         if (data) {
                             data.pop();
-                            that._redrawPad(data).then(
-                                that._updateValue.bind(that)
-                            );
+                            that.promisedRedrawPad = that.promisedRedrawPad
+                                .then(() => that._redrawPad(data))
+                                .then(that._updateValue.bind(that));
                         }
                     })
                     .end()
@@ -381,7 +382,10 @@ class DrawWidget extends Widget {
             .getObjectUrl(file)
             .then(async (objectUrl) => {
                 this.baseImage = await this._getBaseImage(objectUrl);
-                await this._redrawPad();
+                this.promisedRedrawPad = this.promisedRedrawPad.then(() =>
+                    this._redrawPad()
+                );
+                await this.promisedRedrawPad;
             })
             .catch(() => {
                 this._showFeedback(
@@ -518,7 +522,9 @@ class DrawWidget extends Widget {
             this.canvas.width = this.canvas.offsetWidth * ratio;
             this.canvas.height = this.canvas.offsetHeight * ratio;
             this.canvas.getContext('2d').scale(ratio, ratio);
-            this._redrawPad(this.pad.toData());
+            this.promisedRedrawPad = this.promisedRedrawPad.then(() =>
+                this._redrawPad(this.pad.toData())
+            );
         }
     }
 
--- a/packages/enketo-core/src/widget/draw/draw-widget.js
+++ b/packages/enketo-core/src/widget/draw/draw-widget.js
@@ -15,90 +15,6 @@ import downloadUtils from '../../js/download-utils';
 const DELAY = 1500;
 
 /**
- * SignaturePad.prototype.fromDataURL is asynchronous and does not return
- * a Promise. This is a rewrite returning a promise and the objectUrl.
- * In addition it also fixes a bug where a loaded image is stretched to fit
- * the canvas.
- *
- * @function external:SignaturePad#fromObjectURL
- * @param {*} objectUrl - ObjectURL
- * @param {object} options - options
- * @param {number} [options.ratio] - ratio
- * @param {number} [options.width] - width
- * @param {number} [options.height] - height
- * @return {Promise} a promise that resolves with an objectURL
- */
-SignaturePad.prototype.fromObjectURL = function (objectUrl, options) {
-    const image = new Image();
-    options = options || {};
-    const deviceRatio = options.ratio || window.devicePixelRatio || 1;
-    const width = options.width || this._canvas.width / deviceRatio;
-    const height = options.height || this._canvas.height / deviceRatio;
-    const that = this;
-
-    this._reset();
-
-    return new Promise((resolve) => {
-        image.src = objectUrl;
-        image.onload = () => {
-            const imgWidth = image.width;
-            const imgHeight = image.height;
-            const hRatio = width / imgWidth;
-            const vRatio = height / imgHeight;
-            let left;
-            let top;
-
-            if (hRatio < 1 || vRatio < 1) {
-                // if image is bigger than canvas then fit within the canvas
-                const ratio = Math.min(hRatio, vRatio);
-                left = (width - imgWidth * ratio) / 2;
-                top = (height - imgHeight * ratio) / 2;
-                that._ctx.drawImage(
-                    image,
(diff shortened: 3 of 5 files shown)


lognaturel:
Thanks for diving deep! The screencast doesn't seem to work. If you're able to try again with it that would be great.

I'm going to be away for the next 2 weeks and will try to get back to this in mid-August. In the mean time, if someone else is up for review, that would be appreciated.

Am I understanding correctly that edits will still result in further resolution loss if the edit session is opened with a small screen size? I think that's unchanged behavior but want to make sure I understand.

Could you please verify the change with a default image and compare the behavior before and after the change? I think it's unchanged but would be good to double check. I think there's a form in the PR I linked to.

jkuester:
> The screencast doesn't seem to work.

I have noticed that videos (at least some of them) on GitHub do not work with FireFox.  This one does load fine for me on Chrome.  Either way, [here is a direct link](https://drive.google.com/file/d/1paDzi7_sHDZxctaZNgCuSrHqyFPfiVLn/view?usp=sharing) to download the same video.  

> Am I understanding correctly that edits will still result in further resolution loss if the edit session is opened with a small screen size?

Correct.  That is my _expected_ behavior as well.  (However, editing drawings in the CHT is a bit bugged right now, so I have not been able to confirm this exact workflow.)  

> Could you please verify the change with a default image and compare the behavior before and after the change?

As with editing drawings, the CHT does not yet support providing a default image for annotating.  I have not been able to figure out how to attach existing images to a form when running the enekto-core development server.  Is that possible?  Otherwise, it seems like maybe I should be running enketo-express or something.  However, when looking at the Express README is seems like I would also need to be running an OpenRosa server to really do anything useful.  Do you have any pointers on the most simple development/testing setup for enekto-core that would allow for attaching images and editing previously submitted forms?

I was able to test this code with these forms (and things work as expected):
- [drawing-bug3](https://docs.google.com/spreadsheets/d/1WlWhO8D2DAffkGT9EpnVY-lMu6Tmm_M1WDLwtP7kC-A/edit?gid=0#gid=0)
-  [All question types](https://docs.google.com/spreadsheets/d/1af_Sl8A_L8_EULbhRLHVl8OclCfco09Hq2tqb9CslwQ/edit?gid=0#gid=0)

But, I do not have access to [this form](https://docs.google.com/spreadsheets/d/1VaT9ayWE3aOHWGbaHudG3UQ1wL1RMwQtZ0mqGP3VeGg/edit#gid=0) for the default image (have just requested access).

lognaturel:
Apologies for the radio silence here!

I just gave [the default image form](https://docs.google.com/spreadsheets/d/1VaT9ayWE3aOHWGbaHudG3UQ1wL1RMwQtZ0mqGP3VeGg/edit#gid=0) a try with [this image](https://upload.wikimedia.org/wikipedia/commons/c/c7/Saturn_during_Equinox.jpg) as a default and the image is shown significantly scaled down with this PR:

Screenshot 2024-08-13 at 3 27 44 PM

There's the same issue with edits. Here is what it looks like from `main`:
Screenshot 2024-08-13 at 3 29 31 PM

It looks like the download button has also disappeared.

> attach existing images to a form when running the enekto-core development server. Is that possible? 

I don't believe so. I think you'd have to write an implementation for the file management interface.

> pointers on the most simple development/testing setup for enekto-core that would allow for attaching images and editing previously submitted forms?

I believe Martijn usually used the barebones https://github.com/enketo/centro 

I run ODK Central directly from source on my machine. Most folks who do this report it's straightforward to get up and running but there could be issues on different platforms and it is a few different pieces to manage. If you want to try that, you would need postgres, https://github.com/getodk/central-backend/ that you start with `make run` and https://github.com/getodk/central-frontend/ that you start with `npm run dev`. You can then follow instructions for running Enketo Express at https://github.com/getodk/central-frontend/blob/master/docs/enketo.md

jkuester:
I also apologize for the radio silence!  :sweat_smile:   Been swamped the last couple of weeks (amusingly with other widget related work in cht-core...).  

I have no new progress to report yet, but just wanted to let you know that I have not forgotten about this PR and will be circling back this week for a closer look at what you have posted above!

jkuester:
Okay, @lognaturel good news!  I was able to get a proper test environment by [following these steps](https://github.com/enketo/enketo/issues/1349) to spin up enketo-express with an instance of centro. 

With that I was able to recreate (and fix) the behavior your mentioned.  I also thoroughly tested the various permutations of the widget (draw, signature, annotate, annotate with default image) on the first page of a form and last page of a form. in both the normal mode and with `?touch=true`.  

At this point I am feeling pretty confident that I was able to shake out all the weirdness and the widget behavior now matches `main` (in the places that it has not been improved).  Things should be ready for you to have another look!  :pray:

lognaturel:
Thanks so much for doing that follow-up! I should be able to take a deeper look early next week and hopefully get this merged. 🚀

lognaturel:
Thanks again!

Here's what I'm seeing now:
- If I sign (or draw or annotate), resize my browser so only a subset of the signature is shown, and then submit, only the subset of the image is submitted.
- If I annotate a default image, resize my browser so only a subset of the image is shown, submit, enlarge my browser, the default image is shown scaled down. Refreshing the canvas keeps the scaled down image.

Other than that, I agree the bug you were targeting is addressed. I looked at empty states, resizing in other sequences, the test forms I shared and all of that looks good.

jkuester:
@lognaturel, first of all, thank you for the very thorough testing here!  :sweat: The way I was running with centro, I could see that values were being "saved" for the drawings and I assumed that those contained the same image data as what I saw when I used the download button to check the images.  Unfortunately, it turns out this was not the case since the file-manager code was actually scraping the image data from the `canvas` elements (and so it was dependent on the final size of the canvases when the form was saved). 

I tweaked up my centro deployment to actually save the files to disk when the form was submitted and was able to re-create the behavior you described.  :+1: 

Now, for the fix, I am very interested to hear your thoughts on which approach is best.  I went ahead and added the most straightforward and simple "solution" to the branch here. The image data was already being cached in the widget state, but the file-manager could not access it there. So, I updated the widget to actually cache the image data right on the input element in the DOM where the file-manager can read it directly and build the necessary file blob.  This approach keeps things simple and centralized without a lot of moving parts. The downside is that other consumers of enekto-core _with custom file-manager implementations_ may run into the same problem behavior if they had similar canvas-based functionality for loading the image files.  These implementations would need to be updated to read the cache similar to what I did for the default file-manager and the enekto-express file-manager.

Alternatively, it seems _possible_ that perhaps I could maintain the existing canvas-based file-manager behavior by adding a `before-save` hook to the draw widget that would update the canvas size to match the size of the image in the pad so that when the file-manager loads the image from the canvas, it is loading the full image.  The downsides of this include:

- It is more moving parts - consumers need to be triggering the `before-save` event (which apparently the CHT is not currently doing... :facepalm: ) and they need to do this before getting the current files for the form and the triggered canvas resize needs to complete before they get the current files....
- It also is semantically a bit confusing since now the draw widget code treats the canvas only as a display artifact while the actual source of truth for image data is in the pad/cache.  (e.g. the reason I missed this behavior in the first place is because I did not realize that the file-manager was making assumptions about the size and contents of the canvas elements.)

Perhaps there is a 3rd approach I missed that would be preferable?

lognaturel (approved):
I've come back to this a few times to think through the behavior and the implementation. I don't have a better alternative to propose!

We can release a major version for Core and a minor for Enketo. We'll make sure that the release notes for Core explain the change.

Thanks so much for sticking with it, this really wasn't easy!

PR #1336, opened 2024-07-29
Fixed translation cache of element with i18nNumber

### LINKED ISSUE #1337: Problem of multiselect widgets in a repeating group displaying the same selected options
**Describe the bug**
The problem occurred when editing the form.

If the page contains multiple selectpickers with similar i18n keys (for example, in a repeating group), it will show the wrong selected text (even though the select options themselves are correct).

**To Reproduce**

1. Create a form with multilselect in a repeating group.
2. Add some (2 or more) multiselect in form enketo page. Select different options and keep some multiselect have same number of options selected (for example first multiselect has 3 options selected, second multiselect has 2 options selected.
3. Close the form page to save the input.
4. Edit the same form.
5. We will see that the second multiselect will display **3 selected** instead of **2 selected**.

**Expected behavior**
The third multiselect should display **2 selected**

**Screenshots**
This multiselect xform instances
```
<group_nm7ec42  enk:ordinal="1" >
	<multiselect2>a b c</multiselect2>
</group_nm7ec42>
<group_nm7ec42  enk:ordinal="2" >
	<multiselect2>a b</multiselect2>
</group_nm7ec42>
```
will give us
wrong-select
instead of (the right one)
correct-select

-   OS: Windows 11, MacOS
-   Browser: Chrome

**Additional context**
None.

gushil:
Closes #1337 

-   [ ] Saving offline drafts
-   [x] Editing submissions
-   [ ] Form preview
-   [ ] None of the above

Just found the problem in editing submission. I think that is sufficent in this case.

The usage of combination of i18n and i18nNumber as key is the fast way to cache the result.

Yes, other solution could be removing the usage of i18n (and i18nNumber) in selectpicker widget but that will render the cache useless in this case.

None.

No, just usual multiselect in a repeating group.

### DIFF
--- a/packages/enketo-express/public/js/src/module/translator.js
+++ b/packages/enketo-express/public/js/src/module/translator.js
@@ -87,34 +87,44 @@ const localize = (container, lng) => {
         })
         .then(() => {
             list.forEach((el) => {
-                const key = el.dataset.i18n;
+                let key = el.dataset.i18n;
+                // translation key for element with i18nNumber is i18n-i18nNumber
+                if (el.dataset.i18nNumber) {
+                    key = `${key}-${el.dataset.i18nNumber}`;
+                }
                 if (key) {
-                    if (!cache[key]) {
-                        let options = {};
-                        // quick hack for __icon__ replacement
-                        if (el.dataset.i18nIcon) {
-                            options = {
-                                icon: `<span class="icon ${el.dataset.i18nIcon}"> </span>`,
-                                interpolation: { escapeValue: false },
-                            };
-                        }
-                        if (el.dataset.i18nNumber) {
-                            options = {
-                                number: el.dataset.i18nNumber,
-                            };
-                        }
-                        cache[key] = t(key, options);
+                    let options = {};
+                    // quick hack for __icon__ replacement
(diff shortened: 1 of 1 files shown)


packages/enketo-express/public/js/src/module/translator.js
```diff
@@ -87,34 +87,44 @@ const localize = (container, lng) => {
...
         .then(() => {
             list.forEach((el) => {
-                const key = el.dataset.i18n;
+                let key = el.dataset.i18n;
+                // translation key for element with i18nNumber is i18n-i18nNumber
+                if (el.dataset.i18nNumber) {
+                    key = `${key}-${el.dataset.i18nNumber}`;
+                }
                 if (key) {
+                    let options = {};
```
lognaturel, line 96: Why did the option setting move out of the `!cache[key]` block? They only need to be set if there's going to be a new cache entry, right?

gushil: Yes, you are right.

Fixed.

Thanks for the suggestion!

lognaturel (approved):
I have quickly verified that this fixes the issue. It feels low risk.

PR #1341, opened 2024-08-22
Hide date widget fake input when rendering pdf

### LINKED ISSUE #1340: Double date displayed when rendering PDF
**Describe the bug**
When the form date widget contains a value, rendering the widget in PDF give us double date.

**To Reproduce**

1. Create a form contains item with type date.
2. Select some date in date widget so it will have a value.
3. Save the form.
4. Get the form pdf.
5. The date displayed twice in the pdf.

**Expected behavior**
Pdf file will display the date once. 

**Screenshots**
image

-   OS: Windows 11
-   Edge

**Additional context**
None.

gushil:
Closes #1340 

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [x] None of the above : Print PDF

Get PDF and check the PDF result.

The minimal and fast way to achieve the result.

None.

No. Just usual form contains date item.

### DIFF
--- a/packages/enketo-core/src/widget/text-print/text-print.js
+++ b/packages/enketo-core/src/widget/text-print/text-print.js
@@ -36,6 +36,15 @@ class TextPrintWidget extends Widget {
             this.element.after(printElement);
             this.element.classList.add('print-hide');
 
+            // If previous element is a fake input in date widget, hide it as well
+            const previousElement = this.element.previousElementSibling;
+            if (
+                previousElement !== null &&
+                previousElement.classList.contains('date')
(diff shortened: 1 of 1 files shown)


lognaturel:
Did you also try this with regular printing from Enketo Express? I thought that used the same logic and I'm not seeing the problem you identified with it. I'm a little worried that your new logic could be overly aggressive.

lognaturel (approved):
I think this is pretty low-risk because `previousElementSibling` should guarantee only elements within a single widget's definition are being affected.

Issue #892, opened 2020-07-03
config to disable date string detection


MartijnR:
Idea for the future.

Getting `"'2020-10-10' + 1"`, and `". < /path/to/date"`, to work versus requiring all date and datetime _node values_ and _strings_ to be explicitly wrapped in `date()` affects performance negatively. It would be very worthwhile, at least in some mini-ecosystems, to disable the automatic date/datetime string detection with a configuration item. To leverage the performance improvements to the fullest extent we could:

1. detect whether the expression contains any custom XPath (OpenRosa/ODK) functions and if not, send the whole thing straight to the native evaluator. This could be done in Enketo https://github.com/enketo/enketo/issues/170 or in this evaluator.
2. for any expressions that would have to be handled by this evaluator (i.e. containing OpenRosa/ODK function calls), don't detect date strings so  `"'2020-10-10' + 1"` would result in NaN. This would have to be done in this evaluator.

It would even be possible for form designers to enable this faster mode in the form definition.

Issue #170, opened 2019-11-06
Add 'turbo' mode?


MartijnR:
Some thoughts:

* (huge) performance-boost by requiring dates to always be wrapped inside `date()` in expressions
* ~~disable repeat-position-injection (which is a workaround for https://opendatakit.github.io/xforms-spec/#a-big-deviation-with-xforms, and no longer necessary for recently converted xlsforms - this workaround is causing bugs that are not really fixable)~~ - `already done with odk:xforms-version="1.0.0"`

These could be done as separate settings, but both require a tight control of the form definitions, and it seems only applicable to mini ecosystems such as OpenSRP, OpenClinica, Medic Mobile.

PR #1342, opened 2024-08-31
Significant performance improvement using XPath browser native evaluation

jdimeo:
I have been discussing very slow XPath evaluation with @lognaturel via ODK support and I believe I have pinpointed the problem to this call to `calculate()` which never tries native evaluation. In my case, native vs. JS/slow produced the same result, so my understanding is that native should've been attempted and would've worked, and we already have protections like the `OPENROSA` regex to avoid calling the browser with XPath it won't understand. Are there any side effects to this change?

image

Closes #

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [ ] None of the above

### DIFF
--- a/packages/enketo-core/src/js/calculate.js
+++ b/packages/enketo-core/src/js/calculate.js
@@ -344,7 +344,8 @@ export default {
                       newExpr,
                       'string',
                       props.name,
-                      props.index
+                      props.index,
+                      true // Try native XPath evaluation
                   )
                 : '';
 


jdimeo:
More thoughts after mulling on it a bit. We have several options, I'll rank them from most cautious/least invasive to most aggressive, but I would argue the best way forward.

1. What I did- send `true` in this one case
2. Remove the parameter I added so it goes back to the default, but make the default for `tryNative` to `true`. Other callers who omit the parameter will benefit from trying native by default.
3. Remove the parameter altogether. What are the situations in which we would "force" it to only do JS evaluation? 
4. Remove the parameter as well as the `OPENROSA` regex. It seems to me that we can just use the graceful fallback that's already implemented to always and optimistically try native, and if that fails, fall back to slow JS eval. I see that the console statement was already commented out, so we don't have to worry about "console spam".

I think we should do 4 because we saw (now that I properly understand the code) that this has happened in the past... the presence of `position()` was tripping slow JS eval even though it _would've_ worked as native. It's only `position(..)` that was OpenRosa specific. So that's two examples, mine and `position()`, where native _would've_ worked but we artificially constrained it to never try. Not having benchmarked it, I'm guessing that native regex evaluation for `OPENROSA` would be more or less the same as native XPath evaluation to see if it works, but maybe you all have more real world data on that.

lognaturel:
@jdimeo thanks so much for digging into this! Have you tried this change and observed a measurable performance difference? If you have not, I can give it a try.

Unfortunately this part of the code has a lot of vestigial aspects to it. XPathJS is no longer used and was replaced some years ago by https://github.com/enketo/enketo/tree/main/packages/openrosa-xpath-evaluator The intention for that library is to wrap browsers' `evaluate` function, adding support for extensions that Enketo uses. It is my belief that it's now slower to pass in `true` for `tryNative` because of the regex check. I believe that when the evaluator switch happened, that parameter should have been removed. I could very well be wrong and can dig into it more in the coming days. In the mean time, @alxndrsn, maybe you have a sense of whether what I described sounds right?

jdimeo:
@lognaturel while we are self-hosting ODK, we don't have a build environment set up to do any customizations to the codebase- we are using standard images (if I understand correctly) so no, I haven't tried it.

> It is my belief that it's now slower to pass in true for tryNative because of the regex check.

Slower than what? My argument is that the regex check vs. the "always try native and fall back to openrosa-xpath if error" code paths would be roughly equivalent performance. So the regex isn't really saving much pain, it's just avoiding native evaluations that would've worked. But it sounds like you're advocating for 3 in my list above which is also great.

lognaturel:
There's nothing like an empirical check. Your hunch was indeed correct and directly going the straight native route is significantly faster for the form you are looking at. On my machine, it cuts load time from 25s to 3s.

I spoke briefly with @eyelidlessness who reminded me he had considered making this change but backed out of it because it is not guaranteed to work for all expressions. I believe in particular that date processing is an issue (https://github.com/enketo/enketo/issues/892). 

> I believe that when the evaluator switch happened, that parameter should have been removed.

This is wrong for the reasons mentioned above. I think maybe only the comments need to be updated after all.

I now see that the `true` value is only sent for controlled expressions that Enketo relies on, not for user-specified expressions.

@alxndrsn (he's one of the original authors of `openrosa-xpath-evaluator`), does the date processing mean `openrosa-xpath-evaluator` never goes fully native after all? Is that the only blocker to full native?

There's an issue tracking some way to allow users to opt into native evaluation with reduced syntax availability at https://github.com/enketo/enketo/issues/170

jdimeo:
Thanks for digging into this @lognaturel !

I'm not following this:
> but backed out of it because it is not guaranteed to work for all expressions.

If it "doesn't work", we catch the exception and fall back to slow JS expressions that does?

Or are you saying there are date-based edge cases where native *does* "work" (no error) but returns wrong/blank results?

lognaturel:
Some proposed follow-up:
- [ ] for now consider a server-level config key to try native evaluation for user-supplied expressions and document what the ramifications are so that specific deployments can opt-in
- [ ] make it more explicit in https://github.com/enketo/enketo/tree/main/packages/openrosa-xpath-evaluator that native evaluation is never directly used and why
- [ ] update comments that mention XPathJS
- [ ] make it clear that the `tryNative` parameter is not intended to be used for user-supplied expressions and why

eyelidlessness:
I should add that besides strings, numbers can also be interpreted as date/time/datetime values. I believe in Web Forms this is effectively limited to unary/binary (sub) expressions, and casts. I couldn't say offhand if that limitation is the same for ORXE, I'd ask @alxndrsn or have to reference the source to say with certainty.

lognaturel:
> there are date-based edge cases where native does "work" (no error) but returns wrong/blank results? 

Yes, exactly.

alxndrsn:
> I couldn't say offhand if that limitation is the same for ORXE, I'd ask @alxndrsn or have to reference the source to say with certainty.

I think there are some conflicting requirements in the existing implementation around type coercion of various things to & from dates.  Some examples can be seen in the tests, e.g.

* https://github.com/enketo/enketo/blob/89a88e3aa8f0f253f0f8bf3679776342c8c6948a/packages/openrosa-xpath-evaluator/test/integration/openrosa-xpath/date.spec.js#L46
* https://github.com/enketo/enketo/blob/89a88e3aa8f0f253f0f8bf3679776342c8c6948a/packages/openrosa-xpath-evaluator/test/integration/date-comparison.spec.js#L30

> I believe in Web Forms this is effectively limited to unary/binary (sub) expressions, and casts.

It's great to hear the "spec" is being tightened up :+1:

PR #1343, opened 2024-08-31
Add support for Google Analytics 4.

boazsender:
Removes support for Google Analytics 3.

-   [X] Online form submission
-   [X] Form preview

I created a test google analytics 4 property, added it to my enketo-express config.json, and confirmed that events are being recorded.

I copied the way former google analytics 3 implementation worked.

Google analytics 3 config will no longer work. Only a google analytics 4 `measurementId` config property will work.

Any form should work.


lognaturel:
Can you also update https://github.com/enketo/enketo/blob/13616ecf253297465e90a78ae36691ac676e7343/packages/enketo-express/config/sample.env#L53, please?

Strictly speaking I think this would be a semver major change. Because Google Analytics 3 is not supported anymore so it's more like removing broken functionality and then adding something new, I think we can do semver minor, does that sound right?

boazsender:
@lognaturel I've updated the sample env with my guess about how the env file works. I only tested with the config.json. Can you confirm it looks right, or advise me how to test with the env file locally?

Semver minor sounds right to me even though this is theoretically not backward compatible, since GA3 is deprecated, and is not working in folks' instances. Would have left GA3 in if it were operable.

lognaturel:
The approach for converting config file entries to env var names is described at the top of the sample file. What you've done looks good.

You can try it by omitting the config file and setting the desired env variables.

lognaturel (dismissed):

lognaturel:
Could you please look at the failing test?

boazsender:
Test updated. Wow, great to see the amazing test coverage in this project. I'll be sure to run them in the future prior to sending PRs.

lognaturel (approved):

lognaturel:
Yes! It has thorough linting as well so you'll also want to run `yarn lint` or `yarn lint --fix`.

PR #1345, opened 2024-09-03
Fix `bootstrap-datepicker` CSS file import

### LINKED ISSUE #14: SCSS compilation error when consuming enketo-core from Vite project
**Describe the bug**
When consuming the enketo-core SCSS using [Vite](https://vitejs.dev/), I get an import issue from 
https://github.com/enketo/enketo/blob/15e164eabec6d49319eb720e28470ed854f6dfdc/packages/enketo-core/src/widget/date/datepicker-extended.scss#L1

> [vite:css] [sass] Can't find stylesheet to import.

The SCSS file assumes it is always in `/packages/enketo-core` (in the monorepo) but if you obtained the SCSS using `npm install enketo-core --save` then that path does not exist.
Also, the reference in the SCSS file misses the css extension (it ends with `.min`).

**To Reproduce**
1. Create a Vite project using `npm create vite@latest`
2. Add `enketo-core` and `sass` as dependencies to `package.json`
3. Run `npm install` and `npm run dev`
4. Add `<link rel="stylesheet" href="/node_modules/enketo-core/src/sass/plain/plain.scss" />` to your `index.html`

**Expected behavior**
No SCSS compilation error when consuming enketo-core

**Screenshots**
image

sergioestevess:
Closes #14 .

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [x] None of the above

Consuming `enketo-core` v8.3.0 in Vite for both development and production works.

An alternative is "plain `@import`" the minified `bootstrap-datepicker.min.css`, note the extension, but letting the compiler of choice handle the compilation sounds appropriate.

If your compiler is not minifying the SCSS output then we're including a slightly larger CSS module.

No, it shouldn't have an effect on any existing form definition.

### DIFF
--- a/packages/enketo-core/src/widget/date/datepicker-extended.scss
+++ b/packages/enketo-core/src/widget/date/datepicker-extended.scss
@@ -1,4 +1,4 @@
-@import '/packages/enketo-core/node_modules/bootstrap-datepicker/dist/css/bootstrap-datepicker.min';
+@import '/packages/enketo-core/node_modules/bootstrap-datepicker/dist/css/bootstrap-datepicker';
 
 /** fixes by martijn **/
 


lognaturel (dismissed):
Thanks!

lognaturel:
Haven't merged this yet because I do want to do a quick sanity check with EE.

lognaturel (changes requested):
I'm surprised to see that this appears to break the `month` and `year` appearances for dates in EE somehow. See [Annotate default image.xml.zip](https://github.com/user-attachments/files/16948086/Annotate.default.image.xml.zip)

`main`:
Screenshot 2024-09-10 at 9 02 51 AM

this branch:
Screenshot 2024-09-10 at 9 44 16 AM

sergioestevess:
Let me dig a bit more into it, but from a quick look in:

https://github.com/enketo/enketo/blob/461cc32491cee97d3f6963d53576167c3c95cf8c/tools/grunt/sass-paths.js#L7

Looks like Sass imports require a file extension...

sergioestevess:
I believe we can close this – see https://github.com/enketo/enketo/issues/14#issuecomment-2357956755.

Issue #1352, opened 2024-09-11
Images on survey questions not being displayed 


FrankApiyo:
**Describe the bug**
In the choices sheet, a form with images labels as follows fails to render the images as expected:
```
media::image::wolof(wl)
```

**To Reproduce**
Upload the form linked below with the attached media files

[emojis_and_audios.zip](https://github.com/user-attachments/files/16963828/emojis_and_audios.zip)
[EVAL_FEP (1).xlsx](https://github.com/user-attachments/files/16963830/EVAL_FEP.1.xlsx)

**Screenshots**
![Screenshot from 2024-09-11 15-49-20](https://github.com/user-attachments/assets/58bbfc7c-1367-4c97-89d3-42180bbd26a0)

**What's expected**
![Screenshot from 2024-09-11 14-14-16](https://github.com/user-attachments/assets/0b82fa4f-0f57-415a-b407-e10bba8a1abf)

**Browser and OS (please complete the following information):**

-   OS: ubuntu
-   Browser: Chrome

faith-mutua:
This has been resolved in stage.ona.io

PR #1353, opened 2024-09-11
Fix issue with displaying image choices in translated form

FrankApiyo:
Closes https://github.com/enketo/enketo/issues/1352

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Added test cases
Adding `option-label` class to `img`, `audio`, and `video` labels adds styling that might not have been intended
I haven't found a test case that fails or seen an issue that might be considered a regression 
Please find a form attached to the linked issue


packages/enketo-core/src/widget/select-media/select-media.js
```diff
@@ -16,7 +16,7 @@ class MediaPicker extends Widget {
 
     _init() {
         this.element
-            .querySelectorAll('.option-label')
+            .querySelectorAll('.option-label:not(video):not(audio):not(img)')
```
lognaturel, line 19: This is no longer necessary, right?

lognaturel (commented):
Thank you! I think one of your changes is no longer necessary.

packages/enketo-core/src/widget/select-media/select-media.js
FrankApiyo: Right! I'll go ahead and remove

lognaturel (approved):

PR #1364, opened 2024-10-28
enketo-express/api: get server_url and form_id consistently

alxndrsn:


lognaturel:
The context here is that the server URL used to send the API key is always retrieved from the body first and the query param second: https://github.com/enketo/enketo/pull/1364/files#diff-fee8e6db9edd9ba63862ae6175a358a9823518161e41bb0af65cb77a705ccfb8L70

It looks to me like other than that, the GET methods use query parameters only and POST use one or the other. This inconsistency means it's possible to write some degenerate GET requests that send the API key to one server and then make a request to a different one.

Following https://github.com/enketo/enketo/pull/1366, the body isn't read for GET requests anymore so I don't think there's a way to write those bad requests. We also have not been able to identify something truly destructive that can be done with this.

PR #1365, opened 2024-10-29
transformer: only use body-parser for POST endpoint

alxndrsn:


lognaturel (approved):
The `GET` endpoint is intended to be used with a query parameter. There are no other methods used.

PR #1366, opened 2024-10-29
enketo-express: only use body-parser for POST, PUT & DELETE

alxndrsn:


lognaturel (approved):
It's generally not recommended to use a body with `GET` requests. In Enketo, the `GET` endpoints exist primarily for browser-based testing so there isn't an expectation of a body being specified.

PR #1373, opened 2024-11-29
express/translations/es: remove extra clause

alxndrsn:
Closes #

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [x] None of the above

CI.

No.

This should improve the Spanish translation.

No.

### DIFF
--- a/packages/enketo-express/locales/src/es/translation.json
+++ b/packages/enketo-express/locales/src/es/translation.json
@@ -366,7 +366,7 @@
         "export": "Exportar",
         "msg1": "Los registros se almacenan en su navegador hasta que éstos han sido enviados al servidor (incluso aunque apague el ordenador o pierda la conexión).",
         "msg2": "Los registros que están en la cola, excepto aquellos marcados como draft __icon__, se envían al servidor automáticamente en segundo plano, cada 5 minutos cuando la página Web está abierta y hay conexión a Internet.",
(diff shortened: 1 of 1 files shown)


lognaturel:
Can you explain the change? The English string is "Queued records, except those marked as draft __icon__, are uploaded automatically, in the background, every 5 minutes when the web page is open and an Internet connection is available." Is there something wrong with how it's displayed in Spanish?

alxndrsn:
> Can you explain the change? The English string is "Queued records, except those marked as draft **icon**, are uploaded automatically, in the background, every 5 minutes when the web page is open and an Internet connection is available." Is there something wrong with how it's displayed in Spanish?

In English, there are two similar strings: https://github.com/enketo/enketo/blob/11eac8a627e475ad24bc53848f4627797db789ad/packages/enketo-express/locales/src/en/translation.json#L369-L370

In `main` these are currently translated identically in Spanish: https://github.com/enketo/enketo/blob/11eac8a627e475ad24bc53848f4627797db789ad/packages/enketo-express/locales/src/es/translation.json#L368-L369

This PR is to change the second string (`-nodraft`) to remove the text referencing the draft from the Spanish translation ("excepto aquellos marcados como draft __icon__, ").

lognaturel:
Thanks, hadn't seen those two. That's the kind of brief explanation that's super helpful in a PR description. Doesn't even have to be this much detail, just something like "there are two ALMOST identical strings in English that have been translated identically in Spanish."

lognaturel (approved):

PR #1378, opened 2024-12-03
express: restrict test credentials to test envs

alxndrsn:

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [x] None of the above

GitHub Actions

This approach changes minimal amount of code and does not require additional environment variables.

No effect.

No.

### DIFF
--- a/packages/enketo-express/app/models/account-model.js
+++ b/packages/enketo-express/app/models/account-model.js
@@ -41,36 +41,39 @@ function get(survey) {
 
         return Promise.reject(error);
     }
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel:
@alxndrsn discovered that it's possible to make real requests using these test users, particularly in combination with https://github.com/enketo/enketo/pull/1364. There's a small chance someone could be relying on these for something like a health check but really they should not. Ideally these would be defined in test code but for now isolating them seems like a good improvement.

We have not found anything truly destructive that can be done with this.

PR #1379, opened 2024-12-09
Prepare releases

lognaturel:
The changes this covers are https://github.com/enketo/enketo/compare/7.4.0...main

-   [x] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

I did all of the above with the image at https://github.com/enketo/enketo/pkgs/container/enketo/319695763?tag=prep-release

There are currently 3 LTS node versions including 18 (https://nodejs.org/en/about/previous-releases). 18 will go out of maintenance LTS in May of 2025. I briefly considered testing all 3. Given that Node 20 is in maintenance LTS as well, it feels simpler to me to drop 18 now and support the two latest LTS versions.

We've previously done major version changes when changing supported node versions. I don't think we should do this anymore. Here's a good summary of why: https://rtpg.co/2023/09/27/avoiding-major-version-bumps-for-node/

I upgraded specific dependencies involved in security issues or with likely useful improvements since `yarn upgrade` breaks transformer tests.

I briefly explored upgrading eslint to 9 since it's the source of some of our security warnings. Of course, it's only used in development so the security issues don't apply, they just make for some noise. I backed out quickly when I ran into some OOM issues and issues with the cache file. Here's the branch: https://github.com/lognaturel/enketo/tree/eslint9

I think the biggest risk with these releases is around the change to signature/draw/annotate questions. I tried to point to useful resources for folks who use Core with their own file manager but I could imagine that upgrade being confusing. There could also very well be regressions or new undesirable behavior in that area.


yanokwa (approved):
Confirmed version numbers are as I expect

PR #1385, opened 2025-02-08
enketo-express: fix handling for unsupported locales

### LINKED ISSUE #1383: Form crashes with 404 when browser language is set to language that EE doesn't have translations for
**Describe the bug**

**To Reproduce**
1. Set the browser language to a language that Enketo Express doesn't have translations for such as Afrikaans or Malay
2. Open an incognito browser window (necessary for Firefox, otherwise the original browser language is sticky it seems)
3. Go to an Enketo form link

**Expected behavior**
No error, EE falls back to English

**Screenshots**
Image

-   OS: macOS
-   Browser: Firefox 134

**Additional context**
I haven't investigated whether this is a change in behavior or is only being reported by users now because browsers have more locales available.

alxndrsn:
In `i18next` v21, the `whitelist` config option was renamed to `supportedLngs`.

`enketo-express` `i18next` dependency was updated from v20 to v24 in December 2024 in commit 0ac29cbe50de33cd5e0732aaec498817b5c31fa6.

See: https://www.i18next.com/misc/migration-guide#v20.x.x-to-v21.0.0

Closes #1383

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [x] None of the above

Ran enketo-express locally and tested with a weird locale.

It seems straightforward, and inline with `i18next` migration documentation.

Should fix a bug for them.

Should be fixed (and previously broken) for all forms.

### DIFF
--- a/packages/enketo-express/config/express.js
+++ b/packages/enketo-express/config/express.js
@@ -40,7 +40,7 @@ i18next
     .use(I18nextBackend)
     .init({
         // debug: true, // DEBUG
-        whitelist: app.get('languages supported'),
+        supportedLngs: app.get('languages supported'),
         fallbackLng: 'en',
         joinArrays: '\n',
         backend: {
--- a/packages/enketo-express/public/js/src/module/translator.js
+++ b/packages/enketo-express/public/js/src/module/translator.js
@@ -35,7 +35,7 @@ const initialize = new Promise((resolve, reject) => {
         .use(htmlParagraphsPostProcessor)
         .init(
             {
-                whitelist: settings.languagesSupported,
+                supportedLngs: settings.languagesSupported,
                 fallbackLng: 'en',
                 joinArrays: '\n',
                 backend: {


lognaturel (approved):
Thank you! I'm annoyed that I missed that.

alxndrsn:
> Hello. Having tested this, I believe this PR does indeed fix a blocking issue, where browsers with certain locale settings, cannot open the form

Thanks for testing!

> however I believe this PR in effect ignores the browser locale and always defaults to English (so that the user has to select language in the "Choose language" dropdown)

This is possible.  Do you know if browser locale being respected in the version immediately prior to this commit?  I ask because the original bug was caused by an upgrade to various internationalisation dependencies, and it's also possible that the upgrade introduced multiple bugs.

lognaturel:
There are two different translation sources at play here -- the one connected to this error is translations of the Enketo app itself. Following this change, you should observe that when you set the browser to a language that Enketo itself has translations for, Enketo displays text in that language for the submit button and any other system text. If that's not the case, that could be a regression introduced here.

The "Choose language" dropdown shows the form-defined languages. My recollection of how that gets selected is that if there's an explicit `default` set by the form, that takes highest precedent. Then, if the language labels defined in your form include the language IANA subtag in parentheses (https://docs.getodk.org/form-language/), Enketo will switch to the one that matches the browser locale. If that's not the case, it's likely a separate issue.

lognaturel:
I just tried this myself and it does indeed look like it ignores the browser locale. That could likely also explain the form language not being automatically set.

lognaturel:
My apologies, I was running the wrong version. Now I'm seeing the behavior I expect for the Enketo application language.

PR #1387, opened 2025-02-12
Fixed can paste value into read-only date field

### LINKED ISSUE #1386: User can still paste into read-only date field
**Describe the bug**
Users were able to copy and paste values into date fields on forms, even though those fields were set to read-only, which should not have been allowed.

**To Reproduce**

1. Open some form with read-only date field in it.
2. Copy some date value from text editor into clipboard, for example 2024-07-07.
3. Click the read-only date field.
4. Press Ctrl/Cmd+V.
5. Copied value pasted into the read-only date field.

**Expected behavior**
Copied value should not be pasted into the field.

-   OS: Windows 11
-   Browser Edge, Chrome

gushil:
Closes #1386 

-   [ ] Saving offline drafts
-   [x] Editing submissions
-   [x] Form preview
-   [ ] None of the above

Manual test.

Simple yet works.

None.

No, just usual form with read-only date field in it.

### DIFF
--- a/packages/enketo-core/src/widget/date/datepicker-extended.js
+++ b/packages/enketo-core/src/widget/date/datepicker-extended.js
@@ -87,48 +87,45 @@ class DatepickerExtended extends Widget {
     _setChangeHandler($fakeDateI) {
         const { settings } = this;
 
-        let changeEvent = 'change';
         if (!$fakeDateI.closest('label').hasClass('readonly')) {
-            changeEvent += ' paste';
-        }
+            $fakeDateI.on('change paste', (e) => {
+                let convertedValue = '';
+                let value =
+                    e.type === 'paste'
+                        ? getPasteData(e.originalEvent ?? e)
+                        : this.displayedValue;
 
-        $fakeDateI.on(changeEvent, (e) => {
-            let convertedValue = '';
-            let value =
-                e.type === 'paste'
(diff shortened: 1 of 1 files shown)


packages/enketo-core/src/widget/date/datepicker-extended.js
```diff
@@ -87,7 +87,12 @@ class DatepickerExtended extends Widget {
     _setChangeHandler($fakeDateI) {
         const { settings } = this;
 
-        $fakeDateI.on('change paste', (e) => {
+        let changeEvent = 'change';
+        if (!$fakeDateI.closest('label').hasClass('readonly')) {
+            changeEvent += ' paste';
+        }
+
+        $fakeDateI.on(changeEvent, (e) => {
```
lognaturel, line 95: Thanks for noticing this and taking on a fix! This works but feels a bit hard to understand to me. 

No `change` event should ever be fired on a readonly date, right? Could we instead not register the handler at all in that case? Something like 

```
if (!$fakeDateI.closest('label').hasClass('readonly')) {
  $fakeDateI.on('change paste', (e) => {
    ...
}
```

gushil: Got it. Adressed in the next commit.

Thanks.

lognaturel (approved):
Thank you!

PR #1390, opened 2025-02-25
Update test dependencies for new Ubuntu

lognaturel:
Closes #

-   [ ] Saving offline drafts
-   [ ] Editing submissions
-   [ ] Form preview
-   [ ] None of the above


yanokwa (approved):
I don't want to block merging, but would have downgrading to Ubuntu (EOL: April 2025) been a cleaner fix?

I was a little concerned about the no sandbox, but I agree that the risk in an ephemeral CI server is pretty safe

PR #1420, opened 2025-08-07
feat(audioRecording): implement background audio recording submission DEV-213

pauloamorimbr:
### 🗒️ Checklist

1. [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
2. [x] assign yourself
3. [x] fill in the template below and delete template comments
4. [x] update all related docs (README, inline, etc.), if any
5. [x] review thyself: read the diff and repro the preview as written
6. [x] undraft PR & confirm that CI passes
7. [x] request reviewers & improve according to review
8. [ ] delete this section before merging

### 📣 Summary
This PR implements the submission of a recorded background audio in a form.

### 📖 Description
This PR implements:
- New `prepareData` function for all the widgets so the data can be asynchronous prepared by the widget before the form is sent
- New `cleanup` function for all widgets so each instance of the widget can be properly cleaned before removed (as in stopping an audio recording in progress)
- Attaching of the background audio recorded to the submission
- Display of a banner when editing a submission or recovering a saved draft to let the user know that there is a background audio recorded in that submission

I have verified this PR works by manually testing (see CONTRIBUTING.md):

- [x] Online form submission
- [x] Offline form submission
- [x] Saving offline drafts
- [x] Loading offline drafts
- [x] Editing submissions
- [x] Form preview
- [ ] None of the above

### 👀 Preview steps
1. ℹ️ have an account and a project (feel free to use this one with back ground audio setup
[bg-audio-recording.xlsx](https://github.com/user-attachments/files/21719498/bg-audio-recording.xlsx))
2. Setup Enketo to work with KPI ([docs](https://www.notion.so/kobotoolbox/2025-08-11-Frontend-meetup-24c7e515f65480f8976cf77c61514d71)) 
3. Open the form in Enketo and collect data
4. Check the generated data on Data table
5. Save a draft with a background audio and recover the draft
6. Edit a submission
7. All the operations should work properly and Data should represent the one collected.

### DIFF
--- a/packages/enketo-core/src/js/widget.js
+++ b/packages/enketo-core/src/js/widget.js
@@ -99,6 +99,13 @@ class Widget {
     }
 
     /**
+     * Cleans up the widget instance. This is called when the form is reset.
+     */
+    cleanup() {
+        // Part of interface, to be overridden
+    }
+
+    /**
      * Returns widget properties. May need to be extended.
      *
      * @readonly
--- a/packages/enketo-core/src/js/widget.js
+++ b/packages/enketo-core/src/js/widget.js
@@ -187,6 +187,10 @@ class Widget {
         this.element.dispatchEvent(event.Change());
     }
 
+    async prepareData() {
+        return Promise.resolve();
+    }
+
     /**
      * Returns its own name.
      *
--- a/packages/enketo-core/src/js/widget.js
+++ b/packages/enketo-core/src/js/widget.js
@@ -194,6 +194,12 @@ class Widget {
         this.element.dispatchEvent(event.Change());
     }
 
+    /**
+     * Prepares data for the widget. This is called before the form is submitted.
+     * It is meant to be overridden.
+     *
+     * @return {Promise} promise that resolves when the data is prepared
+     */
     async prepareData() {
         return Promise.resolve();
     }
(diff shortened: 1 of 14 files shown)


packages/enketo-core/src/js/widget.js
```diff
@@ -187,6 +194,10 @@ class Widget {
         this.element.dispatchEvent(event.Change());
     }
 
+    async prepareData() {
+        return Promise.resolve();
+    }
```
Akuukis, line 199: why not reuse any of "init"/"setup"/"before"/etc. words usually used as antonyms for the already existing "cleanup"?

Akuukis (changes requested):
Tested, for me the example xls form doesn't trigger backgorund recording, although in the KoboToolbox data table I see an empty background recording column for the submission.

Image

pauloamorimbr:
> Tested, for me the example xls form doesn't trigger backgorund recording, although in the KoboToolbox data table I see an empty background recording column for the submission.
> 
> Image

I just re-tested it and it works as it should for me 🤔 
Try the following please:
- Check if ee.kobo.local has the correct message set for you dev environment, just to make sure its the correct Enketo running
- Try to open the form in Online mode only (not needed, but it forces not to use cache, and it may be the problem)
- Try restarting the Enketo's `yarn watch`. It could be some weird cache.
- Try hard-reloading or deleting the page cache, just to make sure you're not seeing the cached version of the form loaded by some older version.

We can do it together if you want to!

PR #1455, opened 2025-11-10
feat(express): remove pdf generation and prod puppeteer DEV-1250

Akuukis:
### 📣 Summary

Remove a likely unused feature to generate a PDF in order to simplify maintenance.

### 💭 Notes

Notably, as part of that remove puppeteer from production dependencies.

See also https://github.com/enketo/enketo-api-docs/pull/27


pauloamorimbr (approved):
LGTM! 

Well done. Code looks ok
I was able to test previous and PR and it does the work nicely.

PR #1456, opened 2025-11-10
fix(security): sanitize media url parameter and reject attacks DEV-1249

pauloamorimbr:
### 🗒️ Checklist

- [x] assign yourself
    - [x] Online form submission
    - [x] Offline form submission
    - [x] Saving offline drafts
    - [x] Loading offline drafts
    - [x] Editing submissions
    - [x] Form preview
    - [x] None of the above

### 📣 Summary
Implements sanitization and checks on Enketo's media endpoint to avoid path traversal attacks.

### 💭 Notes
**The Security Issue:**  Malicious paths get extracted and passed deeper into the system
Files like `../../../etc/passwd` could be used in cache keys, file operations, or logging

**Potential attack vectors:**

- Cache pollution: Malicious paths could corrupt cache keys
- Log injection: Dangerous paths could appear in logs
- Future vulnerabilities: If code later uses these filenames for file operations
- Information disclosure: Error messages might reveal file system structure

⚠️ **IMPORTANT!**

Even though we recognize the potential risks of the current implementation and the fix is being applied, this vulnerability couldn't be reproduced or explored in any of the theoretically known ways. 
Because of this there isn't a real "before" and "after" scenario to be tested, and the test was done by adding debugging into the code and checking if the mitigation mechanics were acting as expected, but none of those logs are present in this final version of the implementation.

### 👀 Preview steps
1. ℹ️ Run enketo express
2. run 
 ``` bash 
curl -v "http://ee.kobo.local/media/get/1/test/hash/../../../etc/passwd" 2>&1 | grep -E "(Record not present|Not found)"
```
3. 🔴 [on main] The malicious path is processed (checked via logs and debugging. See note above)
4. 🟢 [on PR]  The malicious path is rejected early (checked via logs and debugging. See note above)

### DIFF
--- a/packages/enketo-express/app/lib/media.js
+++ b/packages/enketo-express/app/lib/media.js
@@ -35,8 +35,8 @@ const matchMediaURLSegments = (requestPath) => {
     if (matches != null) {
         const [, resourceType, resourceId, hash, fileName] = matches;
 
-        // Sanitize fileName to prevent path traversal attacks
-        // Use path.normalize and path.basename to ensure only the filename is used
+        // Note: express is already normalizing URL's.
+        // Let's normalize them again for the sake of defense in depth.
         const sanitizedFileName = path.basename(path.normalize(fileName));
 
         // Reject if the normalized path differs significantly from the original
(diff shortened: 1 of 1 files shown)


Akuukis (changes requested):
code LGTM, but can't repro:
1. I get "Not Found" on both `main` and branch, is that ok? I assume it should be "Record not present" in one or other based on the snippet.
2. Please elaborate on the "checked via logs and debugging", as I can't see any relavant logs in the terminal that runs the `yarn develop` locally.

Issue #375, opened 2016-12-09
readonly should be dynamic


MartijnR:
But do not rush to implement this as it has never been requested.

chegeiserCEC:
This feature is still needed. Any update on implementation timeline?

aajeevika-it:
Hello,

Is there an ETA for implementation of this feature in Enketo webforms? It has been in the enketo roadmap for quite a while now

Need this feature very frequently in many kobotoolbox forms. Currently using a workaround by having a separate ‘read only’ version of the question. Apart from not being an elegant(proper) solution, it also leads to overhead in already long and fairly complex forms

thanks

lognaturel:
> check all props.readonly usage in all widgets to remove assumptions that this condition is static

E.g. https://github.com/enketo/enketo/commit/7ac3a8c993e01cf0fb44d3960ab851e155bf0c80 assumes it's static, I believe

PR #1457, opened 2025-11-10
feat: implement dynamic readonly expressions support

iahmedani:
### 📣 Summary

Form fields can now become readonly or editable dynamically based on XPath expressions that evaluate as data changes, just like required, relevant, and calculate fields already work.

### 📖 Description

Previously, fields could only be marked as permanently readonly using `readonly="true()"`. They would stay readonly for the entire session, which limited form flexibility.

Now, readonly status can change dynamically based on conditions in your form. For example:
- A field becomes readonly after a form is approved or locked
- Fields become readonly when certain conditions are met (e.g., when a count reaches a threshold)
- Fields can be conditionally editable based on user roles or status values

This works using XPath expressions like `readonly="./status = 'locked'"` or `readonly="count(../items) > 0"`, matching how required and relevant fields have always worked.

**How this affects users:**
- More flexible forms without workarounds
- Simpler form logic for conditional editing permissions
- Better support for progressive form locking workflows
- Reduced need for complex branching logic to hide/show fields

This implements dynamic readonly expression evaluation following the established architectural pattern used by `required`, `relevant`, and `calculate` attributes.

**Implementation changes:**

1. **Transformer** (`openrosa2html5form.xsl`): Now generates `data-readonly` attribute containing XPath expressions for all readonly bindings
2. **Model layer** (`nodeset.js`): Added `isReadonly(expr)` method that evaluates boolean XPath expressions against the data model
3. **Input layer** (`input.js`): Updated `getReadonly()` to return expression strings from `data-readonly` attribute, with fallback to static readonly check
4. **Readonly module** (`readonly.js`): Rewrote `update()` to evaluate dynamic expressions and apply/remove readonly state based on evaluation results

**Key features:**
- Expression caching for performance (pattern from `required.js`)
- Backward compatible with static `readonly="true()"`
- Participates in the evaluation cascade like other dynamic attributes
- No breaking changes to existing APIs or forms

**XForm example:**
```xml
<bind nodeset="/data/field1" readonly="./status = 'locked'" type="string"/>
<bind nodeset="/data/field2" readonly="count(../items) > 0" type="string"/>
💭 Notes
Closes: #375 (requested since 2016)

Files changed:

packages/enketo-transformer/src/xsl/openrosa2html5form.xsl (+5 lines)
packages/enketo-core/src/js/nodeset.js (+16 lines)
packages/enketo-core/src/js/input.js (+10 lines)
packages/enketo-core/src/js/readonly.js (+59 lines, -7 lines)
packages/enketo-transformer/test/external-fixtures/enketo-core/readonly-dynamic.xml (new test form)
Why this approach:

Follows exact pattern of required, relevant, calculate for consistency
Minimal code changes, maximum reuse of existing infrastructure
Expression caching prevents performance degradation
Regression risk:

Low - maintains full backward compatibility
Static readonly="true()" continues to work identically
Both static readonly attribute and data-readonly attribute are processed
Empty readonly fields continue to receive proper aria-hidden treatment
What's new:

Fields with readonly="false()" now properly become editable (previously treated as static)
Readonly state updates when referenced data changes (via evaluation cascade)
Readonly attribute is now added/removed from DOM elements dynamically
Test form included: readonly-dynamic.xml demonstrates:

Readonly based on status field equality check
Readonly based on numeric comparison
Relative, parent, and absolute XPath patterns
👀 Preview steps
ℹ️ Use the included test form packages/enketo-transformer/test/external-fixtures/enketo-core/readonly-dynamic.xml or create a form with dynamic readonly binding:

<bind nodeset="/data/status" type="string"/>
<bind nodeset="/data/field1" readonly="./status = 'locked'" type="string"/>
Load the form in Enketo

🟢 [on PR] Initially, the status field is empty and field1 is editable (not readonly)

🟢 [on PR] Type "locked" in the status field

🟢 [on PR] Notice field1 immediately becomes readonly (grayed out, not editable)

🟢 [on PR] Change status to anything else (e.g., "unlocked")

🟢 [on PR] Notice field1 becomes editable again

🔴 [on main] The same form would ignore the readonly expression entirely - field1 would remain editable regardless of status value

Additional test case for numeric expressions:

Use binding: <bind nodeset="/data/count" type="int"/> and <bind nodeset="/data/field2" readonly="/data/count > 5" type="string"/>
🟢 [on PR] When count ≤ 5, field2 is editable
🟢 [on PR] When count > 5, field2 becomes readonly
🔴 [on main] field2 would always be editable

### DIFF
--- a/packages/enketo-core/src/js/input.js
+++ b/packages/enketo-core/src/js/input.js
@@ -140,9 +140,17 @@ export default {
     },
     /**
      * @param {Element} control - form control HTML element
-     * @return {boolean} whether element is read only
+     * @return {string|boolean} readonly expression or boolean
      */
     getReadonly(control) {
+        const { readonly } = control.dataset;
+
+        // If data-readonly attribute exists, return the expression
+        if (readonly != null) {
+            return readonly;
+        }
+
+        // Fall back to checking for static readonly attribute (backward compatibility)
         return control.matches('[readonly]');
     },
     /**
--- a/packages/enketo-core/src/js/nodeset.js
+++ b/packages/enketo-core/src/js/nodeset.js
@@ -399,6 +399,22 @@ Nodeset.prototype.isRequired = function (expr) {
 };
 
 /**
+ * @param {string} [expr] - The XPath expression
+ * @return {boolean} Whether node is readonly
+ */
+Nodeset.prototype.isReadonly = function (expr) {
+    return !expr || expr.trim() === 'false()'
+        ? false
+        : expr.trim() === 'true()' ||
+              this.model.evaluate(
+                  expr,
+                  'boolean',
+                  this.originalSelector,
+                  this.index
+              );
+};
+
+/**
  * Validates if requiredness is fulfilled.
  *
  * @param {string} [expr] - The XPath expression
--- a/packages/enketo-core/src/js/readonly.js
+++ b/packages/enketo-core/src/js/readonly.js
@@ -20,14 +20,61 @@ export default {
      * @param {UpdatedDataNodes} [updated] - The object containing info on updated data nodes.
      */
     update(updated) {
-        const nodes = this.form.getRelatedNodes('readonly', '', updated).get();
+        const that = this;
+        const readonlyCache = {};
+
+        // Get nodes with both static readonly attribute and dynamic data-readonly attribute
+        const staticNodes = this.form
+            .getRelatedNodes('readonly', '', updated)
+            .get();
+        const dynamicNodes = this.form
+            .getRelatedNodes('data-readonly', '', updated)
+            .get();
+
+        // Combine both sets of nodes, removing duplicates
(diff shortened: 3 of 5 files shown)


iahmedani:
I have tried to fix this issue using claud ai, please have it review and let's fix this issue once and all.

jnm:
Possibly useful reference:
1. https://github.com/enketo/enketo/issues/375
2. Report that it works in Collect along with example XLSForm: https://github.com/getodk/collect/issues/6149#issuecomment-2406015872
3. Fairly confusing discussion, but likely helpful if we pull in Gareth to review specification compliance or other testing: https://community.kobotoolbox.org/t/setting-required-for-read-only-question-with-calculation/26498/7

pauloamorimbr:
Closed this PR in favor of #1588 that carries this implementation.
Thank you for the contribution, @iahmedani

PR #1466, opened 2025-12-08
ci(releases):  add ee-extra-widget pipeline trigger INFRA-311

olive-KTB:
### 💭 Notes
Automatically trigger the ee-extra-widgets pipeline and add zulip notifications.

### DIFF
--- a/.github/workflows/ghcr.yml
+++ b/.github/workflows/ghcr.yml
@@ -85,7 +85,7 @@ jobs:
           with:
             workflow: build-extra-widgets.yml
             repo: kobotoolbox/enketo-express-extra-widgets
-            ref: cicd
+            ref: main
             token: ${{ steps.app-token.outputs.token }}
             inputs: |
               {


.github/workflows/ghcr.yml
```diff
@@ -1,26 +1,37 @@
-name: GHCR
+name: Build Base Enketo in GCHR and Trigger Extra Widgets
```
Akuukis: we should notify success on zulip as well

.github/workflows/ghcr.yml
```diff
@@ -1,26 +1,37 @@
...
+    build-base:
         runs-on: ubuntu-latest
         permissions:
             contents: read
             packages: write
-
+        outputs:
+          image-tag: ${{ steps.first-tag.outputs.tag }}
+          short-sha: ${{ steps.sha.outputs.short }}
+          branch: ${{ steps.branch.outputs.name }}
```
Akuukis: It's confusing that names don't match, notably `image-tag` vs `first-tag`

.github/workflows/ghcr.yml
```diff
@@ -1,26 +1,37 @@
...
+            - name: Get short SHA
+              id: sha
+              run: echo "short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+      
+            - name: Get branch name
+              id: branch
+              run: echo "name=${GITHUB_REF_NAME}" >> $GITHUB_OUTPUT
+      
             - name: Store version information
               run: git describe --tags > .tag.txt || git rev-parse --short HEAD > .tag.txt
```
Akuukis: FYI latest run has
> fatal: No names found, cannot describe anything.

questions:
1. is that expected behavior?
2. that didn't fail the step because error was in mid of pipeline. Consider using `shell: bash` to apply `-o pipefail` as well. See [SO answer](https://stackoverflow.com/a/75420766)

Akuukis (changes requested):
Good job with linking the two repos, and it works 👌 
- code review: see inline.
- preview: Not tested myself, but I looked through the [latest pipeline](https://github.com/enketo/enketo/actions/runs/20032130002/job/57443772879) that worked fine.
- description: please use the PR template

PR #1485, opened 2026-03-20
fix(enketo-core): show dynamic value in itemset label DEV-2001

jkuester:
### 🗒️ Checklist

- [ ] assign yourself
    - [x] Online form submission
    - [x] Offline form submission
    - [ ] Saving offline drafts
    - [ ] Loading offline drafts
    - [ ] Editing submissions
    - [ ] Form preview
    - [ ] None of the above

### 📣 Summary

Fixes a bug where dynamic values in labels for `select` questions are sometimes not populated.

### 📖 Description

For forms that used dynamic (interpolated) values in labels on the choices sheet (e.g. `Mr. ${name}`), these dynamic values are not populated when rendering the label (e.g. as an option on a `select_one/multiple` question) _unless the form also contained a `repeat`._    

You can reproduce the behavior on https://getodk.org/xlsform/ with a simple form like:

#### survey

type | name | label::en
-- | -- | --
text | name | Name
select_one names | select_name | Select Variant

#### choices 

list_name | name |label::en
-- | -- | --
names | mr | Mr. ${name}
names | mrs | Mrs. ${name}

#### Expected

image

#### Actual

image

The issue seems to be an incomplete implementation of a [this fix](https://github.com/enketo/enketo/commit/add50e136dd59e971bf01dc1b8df625067d7e75a#diff-608210f16bb20aa58380a2606ca9a842e85e4634d9e3d03347845848b9bc0f8c) which is trying to:

> Exclude outputs from the cache, because outputs can be added via itemsets (in labels).

Unfortunately, the ternary [on the next line](https://github.com/enketo/enketo/blob/2aab5ce716effe038fcc66041e4f16dbb908f26d/packages/enketo-core/src/js/form.js#L813) only actually busts the cache if `this.repeatsInitialized` is true.  When a form has dynamic values in the choice labels this produces additional `.or-output` items that are not included in `this.nonRepeats`. In that case we need to skip the cache and query the html directly which is what happens for forms _with a `repeat`_.  I am pretty sure it was just an oversight in the original changes that forms _without a `repeat`_ would still end up just referencing the cache for `.or-output` cases.

This PR simply updates the logic flow here to always skip the cache for `.or-output` (regardless of the `this.repeatsInitialized` value).

Template:
2. do this
3. do that
4. 🔴 [on main] notice that this isn't anywhere BUT it should be here
5. 🟢 [on PR] notice that this is here

### DIFF
--- a/packages/enketo-core/src/js/form.js
+++ b/packages/enketo-core/src/js/form.js
@@ -801,11 +801,15 @@ Form.prototype.getRelatedNodes = function (attr, filter, updated) {
         repeatControls = this.filterRadioCheckSiblings(controls);
     }
 
-    // If a new repeat was created, update the cached collection of all form controls with that attribute
-    // If a repeat was deleted (updated.repeatPath && !updated.cloned), rebuild cache.
-    // Exclude outputs from the cache, because outputs can be added via itemsets (in labels).
-    if (!this.all[attr] || (repeatPath && !cloned) || filter === '.or-output') {
-        // (re)build the cache
+    // Outputs can be dynamically added by itemsets (in labels), so always
+    // re-query the DOM to ensure cloned outputs are included.
+    if (filter === '.or-output') {
+        this.all[attr] = this.filterRadioCheckSiblings([
+            ...this.view.html.querySelectorAll(`[${attr}]`),
+        ]);
+    } else if (!this.all[attr] || (repeatPath && !cloned)) {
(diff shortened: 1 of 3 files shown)


pauloamorimbr:
@jkuester  Thank you very much for your contribution! 🙇🏻

pauloamorimbr (approved):
LGTM! ⭐

PR #1491, opened 2026-03-26
fix(test): add cache model expiration tolerance in tests DEV-1931

pauloamorimbr:
### 💭 Notes
This fixes the cache-model test that keeps failing during CI.
The fail reason is that the test uses a `setTimeout` to check for the time elapsed, but due to nature of `setTimeout` we can have a small variation. That variation, usually 1 ms, is causing the test to fail.
I've added a 10 ms tolerance to the check, so the test will stop failing the CI process.

### 👀 Preview steps
Tests should stop failing frequently due to lack of tolerance in time verification.


duvld (approved):
LGTM

PR #1499, opened 2026-04-08
feat(xml): strip invalid xml characters

pauloamorimbr:
### 📣 Summary

Enketo now strips invalid XML characters when pasting text into form fields, preventing failures when loading drafts or editing submissions that previously contained such characters.

### 📖 Description

When copy-pasting text from sources like poorly formatted PDFs, invisible control characters could end up in form responses. Enketo would accept and submit these characters, but then fail to reload those responses as drafts or for editing. This change ensures invalid XML characters are silently removed on paste, so they never make it into a submission.

### 💭 Notes

- This is a prerequisite for DEV-1129 (rejecting submissions with invalid XML characters on the KPI side). Once this change is deployed, KPI can safely start rejecting such submissions.
- The `document.execCommand('insertText')` approach is used for the paste interception to preserve undo history in the browser.
- 11 unit tests were added covering null chars, control characters, valid whitespace preservation, BMP/SMP ranges, emoji, and realistic bad-PDF scenarios.
Characters outside the [valid XML 1.0 range](https://www.w3.org/TR/REC-xml/#charsets) (`#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]`) are now stripped at three layers in enketo-core:

1. **Paste interception** (`form.js`) — A form-level `paste` event listener on text `<input>` and `<textarea>` elements prevents invalid characters from entering the UI.
2. **`getPasteData()` utility** (`utils.js`) — Widgets using this helper (e.g. the datepicker) automatically receive sanitized clipboard data.
3. **Model-level safety net** (`nodeset.js`) — `Nodeset.prototype.setVal()` strips invalid characters before writing to the XML model, guarding against any input path that bypasses the paste listener.

A new `stripInvalidXmlCharacters()` utility function is exported from `utils.js` for use by integrators if needed.

### 👀 Preview steps

1. ℹ️ have an account and a project with a form containing a text or textarea field
2. copy text containing invisible control characters (e.g. from a badly formatted PDF, or use this command on console: `copy("Hello\x00 from\x01 a\x02 bad\x0B PDF\x0C file\x1F!")`)
3. paste it into a text field in the Enketo form
4. 🔴 [on main] the invalid characters are silently accepted; submitting the form and then trying to edit the submission fails with an XML parse error
5. 🟢 [on PR] the invalid characters are stripped on paste; the submitted data contains only valid XML characters, and editing the submission works correctly

### DIFF
--- a/packages/enketo-core/src/js/form.js
+++ b/packages/enketo-core/src/js/form.js
@@ -7,6 +7,7 @@ import {
     stripQuotes,
     getFilename,
     joinPath,
+    stripInvalidXmlCharacters,
 } from './utils';
 import {
     getXPath,
@@ -1058,6 +1059,29 @@ Form.prototype.setEventHandlers = function () {
 
     // Prevent default submission, e.g. when text field is filled in and Enter key is pressed
     this.view.$.attr('onsubmit', 'return false;');
+
+    // Strip invalid XML characters from pasted text in input fields and textareas
+    this.view.html.addEventListener('paste', (event) => {
+        const el = event.target;
+        if (
+            el.matches &&
+            el.matches(
+                'input:not(.ignore):not([type="file"]), textarea:not(.ignore)'
+            )
+        ) {
+            const pasteData = event.clipboardData
+                ? event.clipboardData.getData('text')
+                : null;
+            if (
+                pasteData &&
+                pasteData !== stripInvalidXmlCharacters(pasteData)
+            ) {
+                event.preventDefault();
+                const sanitized = stripInvalidXmlCharacters(pasteData);
+                document.execCommand('insertText', false, sanitized);
+            }
+        }
+    });
 
     /*
      * The listener below catches both change and change.file events.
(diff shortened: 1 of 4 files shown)


duvld (approved):
Code LGTM, approved because the PR does indeed accomplish what it describes of stripping away bad characters on paste, but I don't think I understand how to reproduce the issue

I couldn't get any submissions through that had bad characters either on a fresh form or editing a submission. It seems like kpi is already rejecting bad xml on submission:
Image

I'm not sure what is meant then by "silently accepted" when submitting a form with bad characters seems to be impossible

pauloamorimbr:
> Code LGTM, approved because the PR does indeed accomplish what it describes of stripping away bad characters on paste, but I don't think I understand how to reproduce the issue
> 
> I couldn't get any submissions through that had bad characters either on a fresh form or editing a submission. It seems like kpi is already rejecting bad xml on submission: Image
> 
> I'm not sure what is meant then by "silently accepted" when submitting a form with bad characters seems to be impossible

This was not KPI denying the submission, but Enketo building a malformed XML for submission due to the invalid XML characters. This fix also addresses that by stripping chars not only on paste but also on nodeset's setVal.

Thank you for documenting this! 🙇🏻

PR #1501, opened 2026-04-10
fix(FormModel): prevent session creation if model is invalid DEV-1945

pauloamorimbr:
### 📣 Summary

Forms with a field named `model` no longer break the session context used for preloads like `username`, `deviceid`, etc.

### 💭 Notes

When a form contained a field named `model`, `querySelector('model')` in `createSession` would match that data field instead of the actual `<model>` root element. This caused the session instance — which populates preload values like `username`, `deviceid`, `phonenumber`, etc. — to be appended to the wrong node, corrupting the form instance.

The fix adds a guard to ensure the matched element is actually the document root (`=== this.xml.documentElement`) before proceeding. This also correctly prevents `createSession` from running on the `FormModel` clone used in `getDataStrWithoutIrrelevantNodes`, which is initialized from instance data (not model XML) and does not need a session.

- `createSession` is called once from `init`, but `FormModel` can be instantiated a second time as a clone via `new FormModel(this.model.getStr())` in `Form.prototype.getDataStrWithoutIrrelevantNodes`. In that case, `modelStr` is the serialized primary instance child (e.g. `<data>...</data>`), not the full model XML — so `documentElement` is never `<model>` and the guard correctly skips session creation.
- The transformer (`openrosa2xmlmodel.xsl` via `docToString`) always produces `<model>...</model>` as the root of `modelStr` for a real form, so the guard never incorrectly blocks the legitimate path.

### 👀 Preview steps

1. Create a form with a field named `model`.
2. Load the form.
3. On main: Notice that the submission fails with an error
4. On branch: Verify that the form submits successfully.

### DIFF
--- a/packages/enketo-core/src/js/form-model.js
+++ b/packages/enketo-core/src/js/form-model.js
@@ -259,6 +259,10 @@ FormModel.prototype.createSession = function (id, sessObj) {
     let instance;
     let session;
     const model = this.xml.querySelector('model');
+    if (!model || model !== this.xml.documentElement) {
+        return;
+    }
+
     const fixedProps = [
         'deviceid',
         'username',
@@ -267,9 +271,6 @@ FormModel.prototype.createSession = function (id, sessObj) {
         'simserial',
         'subscriberid',
     ];
-    if (!model) {
-        return;
-    }
 
     sessObj = typeof sessObj === 'object' ? sessObj : {};
     instance = model.querySelector(`instance#${CSS.escape(id)}`);


duvld (approved):
LGTM

alxndrsn:
> `querySelector('model')` in `createSession` would match that data field instead of the actual `<model>` root element.

This is surprising.  From MDN:

> The [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) method querySelector() returns the first [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) within the document that matches the specified [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors)
>- _https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector_

I'm imagining a scenario like:

```xml
<model>
  <instance>
    <data id="Simple">
      <model/>
      <meta>
        <instanceID/>
      </meta>
    </data>
  </instance>
</model>
```

With this XML, the following returns `true` in Firefox & Chrome:

```js
s = `
  <model>
    <instance>
      <data id="Simple">
        <model/>
        <meta>
          <instanceID/>
        </meta>
      </data>
    </instance>
  </model>
`;
xml = new DOMParser().parseFromString(s, 'text/xml');
xml.querySelector('model')===xml.documentElement;
```

alxndrsn:
:+1: I've understood why this change works:

`new FormModel(...)` is called on two subtley different inputs:

1. on load, with a `<model>...</model>`: https://github.com/enketo/enketo/blob/28f4f916b3fecd76ca0aea383761596e830c0be2/packages/enketo-core/src/js/form.js#L98
2. on submission, with the inner `<data>...</data>`: https://github.com/enketo/enketo/blob/28f4f916b3fecd76ca0aea383761596e830c0be2/packages/enketo-core/src/js/form.js#L945

That's surprising in itself.

Perhaps a more natural check would be:

```diff
-    const model = this.xml.querySelector('model');
-    if (!model || model !== this.xml.documentElement) {
+    if (this.xml.documentElement.tagName !== 'model') {
         return;
     }
```

pauloamorimbr:
> 👍 I've understood why this change works:
> 
> `new FormModel(...)` is called on two subtley different inputs:
> 
> 1. on load, with a `<model>...</model>`: https://github.com/enketo/enketo/blob/28f4f916b3fecd76ca0aea383761596e830c0be2/packages/enketo-core/src/js/form.js#L98
> 2. on submission, with the inner `<data>...</data>`: https://github.com/enketo/enketo/blob/28f4f916b3fecd76ca0aea383761596e830c0be2/packages/enketo-core/src/js/form.js#L945
> 
> That's surprising in itself.
> 
> Perhaps a more natural check would be:
> 
> ```diff
> -    const model = this.xml.querySelector('model');
> -    if (!model || model !== this.xml.documentElement) {
> +    if (this.xml.documentElement.tagName !== 'model') {
>          return;
>      }
> ```

Exactly. It's being called twice in different contexts.

And I understand what you're pointing at. Since there's also the case where a `model` tag was simply not found we would have yet another validation just below that first one, so we just added both validations in one.

PR #1504, opened 2026-04-17
feat(singleSubmission): add full-screen custom submission messages DEV-1114

pauloamorimbr:
### 📣 Summary

Forms can now display a server-provided confirmation message after successful single-submission, rendered as Markdown in a full-screen modal.

### 📖 Description

When a form's model contains a `<submitMessage>` element and an OpenRosa server responds to a submission with a `<message>` element inside `<OpenRosaResponse>`, Enketo now displays that message to the respondent as a full-screen, non-dismissible modal instead of the default redirect/success dialog.

The message supports Markdown formatting (rendered client-side via `markdown-it`). Only single-submission forms (`/single/` URL) are eligible.

### 💭 Notes

- Re-implementation of #1368 (`boazsender-custom-confirmation-message`). That PR was authored by @boazsender but was never merged. This implementation takes a cleaner approach: `markdown-it` is used directly (the original PR used `jstransformer-markdown-it` which is Node.js-only and would not have bundled correctly for the browser), and the `<submitMessage>` opt-in check is isolated in a `getSubmitMessage()` helper rather than threaded as a parameter through the submit stack.
- `connection.js` was refactored to always parse the OpenRosa XML response body — not only on errors — so `result.message` is available on success (201/202) as well.
- A new `gui.fullScreenAlert(message, heading, level, dismissible)` function is exposed for potential reuse.
- The full-screen modal uses `vex-js` with `escapeButtonCloses: false` and no buttons by default, making it truly non-dismissible. A `dismissible` parameter exists for future use.
- The `.vex-dialog-title` is visually hidden but available to assistive technologies (sr-only pattern).
- Regression risk: the `connection.js` change means `response.text()` is now always consumed on upload responses. This should be safe since the body is only read once, but any downstream fork that relied on the response body being unconsumed on success paths should be aware.

The feature requires two conditions to be met:

1. **Client-side opt-in**: the form's model XML must contain a `<submitMessage>` element. Forms without this element continue to behave as before.
2. **Server response**: the OpenRosa server must return a non-empty `<OpenRosaResponse><message>` body on a successful (201/202) submission response.

### 👀 Preview steps

1. ℹ️ Have a form prepared for submission custom  response (e.g: 
[SubmitMessage-Basic.xlsx](https://github.com/user-attachments/files/26841070/SubmitMessage-Basic.xlsx)
[SubmitMessage-jrchoicename.xlsx](https://github.com/user-attachments/files/26841072/SubmitMessage-jrchoicename.xlsx)
[SubmitMessageLocalized.xlsx](https://github.com/user-attachments/files/26841073/SubmitMessageLocalized.xlsx))
2. Open a form in online mode and `single` or `once per responder` mode
3. Fill out and submit the form.
4. 🔴 [on main] After submission, the standard redirect/success dialog appears.
5. 🟢 [on PR] After submission, a full-screen modal displays the server's response message rendered as Markdown.
6. Verify the modal cannot be dismissed via button, close icon, or ESC key.

8. ℹ️ Submit a form **without** `<submitMessage>` in the model — confirm the standard behavior is unchanged.

### DIFF
--- a/packages/enketo-express/app/views/styles/common.scss
+++ b/packages/enketo-express/app/views/styles/common.scss
@@ -67,3 +67,36 @@ progress {
 .log {
     word-break: break-all;
 }
+
+.vex.vex-theme-plain.full-screen {
+    padding: 0;
+
+    .vex-content {
+        width: 100vw; // fallback for browsers without dvw support
+        width: 100dvw; // dynamic viewport: accounts for mobile browser UI
+        min-height: 100vh; // fallback for browsers without dvh support
+        min-height: 100dvh; // dynamic viewport: accounts for mobile browser UI
+        border: none;
+        border-radius: 0;
+    }
+
+    .vex-dialog-message {
+        border: 7px solid #ddd;
+        border-radius: 7px;
+        padding: 1em 3em;
+        margin: 1em 3em;
+    }
+
+    .vex-dialog-title {
+        // visually hidden but available to assistive technologies
+        position: absolute;
+        width: 1px;
+        height: 1px;
+        padding: 0;
+        margin: -1px;
+        overflow: hidden;
+        clip: rect(0, 0, 0, 0);
+        white-space: nowrap;
+        border-width: 0;
+    }
+}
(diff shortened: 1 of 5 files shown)


duvld (approved):
Preview steps work, LGTM

PR #1507, opened 2026-04-24
feat(gui): restyle custom confirmation message  DEV-2019

pauloamorimbr:
### 📣 Summary

When a form defines a custom submission message, it now appears inline within the form page instead of in a full-screen overlay.

### 💭 Notes

- Forms that include a `kobo:submitMessage` attribute in their model can define a custom message to show after a successful submission. Previously, this message was displayed in a full-screen modal dialog (using `fullScreenAlert`). With this change, the message is rendered inline — the form content is replaced with the message in place, while the page header and footer remain visible.
- Footer action buttons (Submit, Save Draft, etc.) are removed once the message is shown, since there is nothing more to do.

### 👀 Preview steps

1. ℹ️ Have a form with `kobo:submitMessage` configures. e.g: [SubmitMessage-Basic.xlsx](https://github.com/user-attachments/files/27061604/SubmitMessage-Basic.xlsx)
2. Open the form in single-submission mode.
3. Fill in and submit the form.
4. 🔴 [on main] Notice a full-screen overlay appears with the custom message, blocking the entire page.
5. 🟢 [on PR] Notice the form content is replaced by the custom message inline, with the form header and "powered by" footer still visible and the action buttons gone.

This form: [SubmitMessageMultilineText.xlsx](https://github.com/user-attachments/files/27061639/SubmitMessageMultilineText.xlsx) can also be good for testing, since it uses a multiline text box as the input for the confirmation message, allowing for testing long markdown formatted strings, e.g: 

```
#### Thank you for your submission!

Your response has been recorded. A member of our team will review it shortly.

If you have any questions, please contact [support@example.org](mailto:support@example.org).
```

Screenshot from 2026-04-24 13-00-37

### DIFF
--- a/packages/enketo-express/public/js/src/module/controller-webform.js
+++ b/packages/enketo-express/public/js/src/module/controller-webform.js
@@ -21,7 +21,6 @@ import encryptor from './encryptor';
 import formCache from './form-cache';
 import { getLastSavedRecord, populateLastSavedInstances } from './last-saved';
 import { replaceMediaSources, replaceModelMediaSources } from './media';
-import markdownit from 'markdown-it';
 
 /**
  * @typedef {import('../../../../app/models/survey-model').SurveyObject} Survey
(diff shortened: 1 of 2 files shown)


duvld (commented):
Do we still want to keep the `fullScreenAlert` component(?) you created for the original fix? It doesn't seem to be used anywhere else

duvld (approved):
LGTM!

PR #1510, opened 2026-05-04
feat(audio): add timestamp postfix to uploaded audio filenames DEV-2070

pauloamorimbr:
### 📣 Summary
This PR adds a postfix timestamp in file names for audio file uploads. That is required to avoid attachment issues.

### 💭 Notes
Currently when an audio file is uploaded its name is used as is for the attachment.
If multiple files with the same name are uploaded to the same form for different questions, then we have a data loss situation and only the last uploaded file is attached and posted.
This fix adds a timestamp to filenames to avoid this issue

### 👀 Preview steps
1. ℹ️ have an account and a project with multiple audio questions
2. Deploy and open the form on Enketo
3. Upload the same file or files with the same name to more than 1 audio question
4. 🔴 [on main] notice that there's only 1 attached audio file and, on KPI, it's not possible to work with transcriptions.
5. 🟢 [on PR] notice that there are multiple files attached with timestamped names and it is possible to work with transcriptions on KPI

### DIFF
--- a/packages/enketo-core/Gruntfile.js
+++ b/packages/enketo-core/Gruntfile.js
@@ -23,6 +23,7 @@ module.exports = (grunt) => {
         client: {
             mocha: {
                 timeout: Number.MAX_SAFE_INTEGER,
+                grep: grunt.option('grep') || undefined,
             },
         },
         reporters: ['dots'],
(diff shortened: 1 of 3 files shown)


jamesrkiger (commented):
After discussion, we decided Paulo would look into adding spec tests if they can be done quickly enough to merge tomorrow.

jamesrkiger (approved):
lgtm

PR #1518, opened 2026-05-22
 feat(svg): improve SVG sanitization using DOMPurify DEV-2163

pauloamorimbr:
### 📣 Summary
SVG sanitization is now improved by using DOMPurify with some specific rules

### 💭 Notes
- Some specific XSS atacks were still possible with the current implementation so, to improve security, we opted to use `DOMPurify` library.
- DOMPurify itself does not handle CSS declarations or properties, but it does provide an [example](https://github.com/cure53/DOMPurify/blob/main/demos/hooks-sanitize-css-demo.html) using hooks that was used as the basis for this implementation.
- To deal with parsing the styles I added `css-tree`, which is a fast and well established CSS parser to avoid dangerous text operations to navigate a CSS tree
- The list of safe properties was researched from various fonts.
- Besides some specific properties, the values are also being sanitized to avoid the use of `url(...)` in property values
- A test form and SVG were added to serve as reference for various possible exploits. It can be checked locally at http://localhost:8005/preview?form=http://localhost:3000/malicious-svg.xml (check the form server port, it may be different)
- Only class declarations or allowed SVG tags are allowed in style definition (so no `body`, `:root` or any other out of the SVG scope is allowed)
- href and url links are sanitized to avoid external content loading/tracking
- All the main common tags and styles are allowed

### 👀 Preview steps
1. ℹ️ load a form that loads malicious SVG. 
4. 🔴 [on main] notice that malicious code is working
5. 🟢 [on PR] notice that element was sanitized and malicious code is not present in page


alxndrsn:
> ℹ️ load a form that loads malicious SVG. e.g: http://ee.kobo.local/preview?form=https://www.alxndrsn.com/example-xforms/image-map-injection/image-map.xform.xml

Hi @pauloamorimbr, I've removed the example form at this URL as I think it's helpful to keep the exploit private until the problem has been fixed, and enketo deployers have had time to update their versions running in production.

pauloamorimbr:
> > ℹ️ load a form that loads malicious SVG. e.g: http://ee.kobo.local/preview?form=https://www.alxndrsn.com/example-xforms/image-map-injection/image-map.xform.xml
> 
> Hi @pauloamorimbr, I've removed the example form at this URL as I think it's helpful to keep the exploit private until the problem has been fixed, and enketo deployers have had time to update their versions running in production.

Hi, @alxndrsn. Ok Thank you.

packages/enketo-core/src/js/utils.js
```diff
@@ -289,74 +291,57 @@ function encodeHtmlEntities(text) {
...
-        });
-    });
-
-    return sanitizedSvg;
+    // Strip <foreignObject> subtrees at the XML level before any HTML parsing
+    // occurs. The onerror payload is dispatched during DOMPurify's internal
+    // HTML parse step.
+    // Doing this pre-strip here ensures dangerous HTML descendants from
+    // foreignObject never reach that parse step.
+    Array.from(
```
jamesrkiger: What would you think of something like this?

```
const inertDoc = document.implementation.createHTMLDocument('');
const container = inertDoc.createElement('div');
container.appendChild(inertDoc.importNode(svgElement, true));
const fragment = DOMPurify.sanitize(
    container.innerHTML,
    DOMPURIFY_SVG_CONFIG
);
return fragment.querySelector('svg');
```

It seems to work locally in avoiding that dialog thrown by the malicious-test svg. The specificity of stripping foreignObject by name is just kind of a red flag to me that we're not catching the underlying issue. And whether or not this is a good solution, I think the fact that it avoids the dialog suggests the problem itself isn't with DOMPurify. Let me know what you think.

packages/enketo-core/src/js/utils.js
```diff
@@ -289,74 +291,57 @@ function encodeHtmlEntities(text) {
...
+    RETURN_DOM_FRAGMENT: true,
+    FORBID_TAGS: [
+        'foreignObject',
+        'iframe',
+        'object',
+        'embed',
+        'link',
+        'meta',
+        'script',
+        'style',
```
jamesrkiger: I think we said we were going to skip the style restriction when we talked with John.

jamesrkiger (changes requested):
Left a couple of comments

jamesrkiger:
@lognaturel  @alxndrsn We have been having some discussion about just how restrictive we should be on style tags and attributes in SVGs. As it stands now (setting aside the current PR), we forbid style tags (i.e. `<style>) on SVGs, but we allow style attributes (i.e. `style="..."`). This was [implemented](https://github.com/enketo/enketo/pull/1451) last year. This policy on its own is fairly restrictive from a user perspective. From my research, it seems like a lot of popular tools for generating SVGs don't prioritize these kinds of security concerns and focus instead on optimization (and a style tag with classes can make sense from that perspective). So we could end up stripping important styling from an SVG and the average user may have no idea why. 

That being said, it seems we have three options here:

1. Block style tags and attributes altogether: A very restrictive approach that may deform or break entirely harmless SVGs.

2. Continue with the current policy of blocking style tags but allowing attributes: A little less restrictive, but with the benefit of being our effective current policy. At the same time, we should probably be clearer about this in documentation going forward.

3. Allow both tags and attributes, but create an allow list of safe CSS attributes and block all others. There are lots of CSS attributes, but most of them are irrelevant for legitimate SVG purposes so we could probably make a manageable list.

I'm increasingly leaning towards option 3 here, but I would be interested to hear your thoughts on this.

jnm:
@jamesrkiger @pauloamorimbr, let's see about option 3 in the absence of any objection, although if having a safe list of CSS attributes is going to delay the overall work of using DOMPurify significantly, we should get this PR merged—even if it allows all CSS—as an incremental improvement and then deal with CSS subsequently.

lognaturel:
Agreed with CSS whitelisting and with @jnm's suggestion to layer it on if needed.

pauloamorimbr:
@copilot resolve the merge conflicts in this pull request

jamesrkiger (changes requested):
@pauloamorimbr Thanks for trying to work in the more nuanced approach to styles here. But I see some potential issues with the implementation (we can talk about it later). In line with John's request to get this merged quickly, can you split off this new work from the basic DOMPurify solution you had added earlier? Then we can get the DOMPurify changes merged and revisit the style issue later.

packages/enketo-core/src/js/sanitization-utils.js
```diff
@@ -0,0 +1,483 @@
...
+    'femergenode',
+    'femorphology',
+    'feoffset',
+    'fepointlight',
+    'fespecularlighting',
+    'fespotlight',
+    'fetile',
+    'feturbulence',
+    'filter',
+    'foreignobject',
```
jamesrkiger: Should probably drop this for consistency since we're blocking this tag elsewhere, right?

pauloamorimbr: Yes. Makes sense!

packages/enketo-core/src/js/sanitization-utils.js
```diff
@@ -0,0 +1,483 @@
...
+            } else {
+                node.removeAttribute('style');
+            }
+        }
+    });
+
+    try {
+        DOMPurify.sanitize(sanitizedSvg, DOMPURIFY_SVG_CONFIG);
+    } finally {
+        DOMPurify.removeHook('uponSanitizeElement');
```
jamesrkiger: A note from my local LLM. Not something I would have noticed myself:
DOMPurify.removeHook(event) without a function reference calls Array.pop(), removing the most-recently-added hook rather than the specific one registered here. DOMPurify v3.4.7 supports removeHook(event, fn) with a function reference. Capturing the closures in named variables and passing them to removeHook would make this robust against any future co-use of DOMPurify hooks elsewhere in the codebase.

packages/enketo-core/src/js/utils.js
```diff
@@ -4,6 +4,8 @@
  * @module utils
  */
 
+import { sanitizeSvg } from './sanitization-utils';
```
jamesrkiger: Do we need to import this here? Maybe you're sticking to an established pattern, but it seems like consumers should just import sanitizeSvg directly.

packages/enketo-core/src/js/sanitization-utils.js
```diff
@@ -0,0 +1,483 @@
...
+
+/**
+ * Sanitizes an SVG element using DOMPurify, removing potentially dangerous
+ * elements and attributes (scripts, event handlers, javascript: URLs, etc.).
+ * @param {SVGElement} svgElement - The SVG element to sanitize
+ * @return {SVGElement|null|undefined} The sanitized SVG element
+ */
+function sanitizeSvg(svgElement) {
+    if (!svgElement) {
+        return svgElement;
```
jamesrkiger: Is the null | undefined distinction important here? If not, I think returning null (or undefined, whichever) would be clearer.

packages/enketo-core/src/js/sanitization-utils.js
```diff
@@ -0,0 +1,483 @@
...
+    };
+}
+
+function normalizeTypeSelectorName(name) {
+    const unprefixedName = name.includes('|') ? name.split('|').pop() : name;
+
+    return unprefixedName.toLowerCase();
+}
+
+function isAllowedSvgSelector(prelude) {
```
jamesrkiger: I think the `prelude` term comes from css-tree, right? Might be nice to explain what it means in this context for future readability.

jamesrkiger (approved):
I left some mostly nitpicky comments. Take a look and adjust as needed, but good to merge after that.

packages/enketo-core/src/js/utils.js
pauloamorimbr: The idea was to keep compatibility in the places it's already being used, but I'll just change the imports there. It's only 2 places.

PR #1532, opened 2026-06-04
fix(security): ensure default params can't target unexpected form elements DEV-2136

### LINKED ISSUE #1508: Query parameter defaults can be used to set values at any XPath path
For example, add `?d[/data/@id]=myTargetProject&d[/data/@version]=7` to the end of an Enketo URL to force submission to a form with formID `myTargetProject`. This could potentially be used to hijack sensitive data entered into a legitimate form.

In general, it feels unexpected that any XPath path is allowed. For example, you can add `?d[/model]=bad` to break the form.

jamesrkiger:
### 📣 Summary

Improves security during parsing of defaults via query params, ensuring that defaults only alter form questions rather than attributes of the form or other unintended targets.

### 👀 Preview steps

1. Create and deploy a form
2. Load form in enketo with valid defaults in query params: `ee.kobo.local/single/<enketo_id>?d[my_question]=ananswer`
3. See that form initializes with intended default value
4. Alter query params to target invalid elements (e.g. `d[/form_id/@id]=someotherid`)
5. Submit form
6. 🔴 [on main] Observe that invalid target was succesfully altered in request  
7. 🟢 [on PR] Verify that invalid target is not altered

### DIFF
--- a/packages/enketo-express/test/client/feature/url-defaults.spec.js
+++ b/packages/enketo-express/test/client/feature/url-defaults.spec.js
@@ -15,7 +15,7 @@ const MODEL = `<model>
   </instance>
 </model>`;
 
-describe.only('URL defaults (_prepareInstance)', () => {
+describe('URL defaults (_prepareInstance)', () => {
     const parser = new DOMParser();
 
     it('applies a default value to a matching leaf node', () => {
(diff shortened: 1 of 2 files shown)


packages/enketo-express/test/client/feature/url-defaults.spec.js
```diff
@@ -0,0 +1,112 @@
...
+      <respondent>
+        <first_name/>
+        <last_name/>
+      </respondent>
+      <meta><instanceID/></meta>
+    </myform>
+  </instance>
+</model>`;
+
+describe.only('URL defaults (_prepareInstance)', () => {
```
pauloamorimbr: I didn't know about this `only` keyword. I researched it and it seems like this could make some runners to only execute this test. Everywhere else we're just using `describe()`, maybe we should stick to it.

pauloamorimbr (approved):
LGTM! 

Left only 1 nitpick comment.

PR #1536, opened 2026-06-11
fix(audioWidget): audio widget buttons being triggered by enter DEV-2273

pauloamorimbr:
### 📣 Summary

Pressing `Enter` on a text input field when an audio widget is present won't trigger the audio recording anymore.

### 💭 Notes

Button elements in the audio widget were missing `type="button"`, causing them to default to `type="submit"`. When Enter was pressed inside an `<input>`, the browser fired an implicit form submission and activated the first submit-type button in the DOM — which happened to be the audio widget's **Start Recording** button. 
The fix adds `type="button"` to all buttons of audio widget to avoid the issue.

### 👀 Preview steps

Template:

1. ℹ️ have an account and a project with an audio widget and other widgets containing text inputs
2. open the form on Enketo
3. Give focus to a text input element and press `Enter`
4. 🔴 [on main] notice that audio widget starts recording
5. 🟢 [on PR] notice that nothing happens other than the expected for the given text input (e.g: search result if in a seartch input)

### DIFF
--- a/packages/enketo-core/src/widget/audio/audio.js
+++ b/packages/enketo-core/src/widget/audio/audio.js
@@ -170,12 +170,12 @@ class AudioWidget extends Widget {
         const stepFragment = document.createRange().createContextualFragment(
             `<div class="step-action-select">
                 <div class="button-group">
-                    <button class="btn-record btn btn-primary small">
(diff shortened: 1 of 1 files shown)


jamesrkiger (approved):
lgtm

PR #1538, opened 2026-06-15
feat(encryption): support editing instances from encrypted forms

### LINKED ISSUE #1583: Support editing submissions from encrypted forms while keeping existing attachments
**Is your feature request related to a problem? Please describe.**
Editing a submission from an encrypted form currently fails to save when the submission contains media attachments. When a submission is loaded for editing, only a preview of the attachments is rendered and the underlying blobs are not refetched. This is fine for unencrypted forms, but encrypted forms must re-encrypt every media file locally before saving, which requires the actual blob rather than just the filename. As a result, saving an edited submission from an encrypted form fails with `TypeError: Require Blob` unless the user manually reattaches every media file.

**Describe the solution you'd like**
When editing an instance of an encrypted form, retrieve the actual media file blobs for existing attachments so the submission can be re-encrypted and saved with the existing photos and attachments kept intact. The behavior should be additive and gated to encrypted forms only, so non-encrypted edit flows keep the existing filename-string behavior and are unaffected.

**Describe alternatives you've considered**
Requiring users to manually reattach every media file each time they edit a submission from an encrypted form. This is error-prone and a poor user experience, especially for submissions with many attachments.

**Additional context**
This was encountered while building a managed form encryption and decryption feature, where a user editing a decrypted submission on Enketo would hit the save failure described above. Implemented in #1538.

kelvin-muchiri:
### 📣 Summary

Adds the ability to edit and save submissions from encrypted forms while keeping existing photos and attachments

### 📖 Description

When a submission is loaded for editing, only a preview of the attachments is rendered; the blobs aren't refetched. This works fine for unencrypted forms, but for encrypted forms, every media file has to be re-encrypted locally before saving since the encryption process needs the actual blob, not just the filename.

We hit this while building our managed form encryption and decryption feature. A user editing a decrypted submission on Enketo would get a save failure with the error `TypeError: Require Blob` unless they manually reattached every media file.

The change is additive and gated: non-encrypted edit flows are unaffected (falls back to the existing filename-string behavior).

closes #1583

### DIFF
--- a/packages/enketo-express/public/js/src/module/controller-webform.js
+++ b/packages/enketo-express/public/js/src/module/controller-webform.js
@@ -134,6 +134,9 @@ function init(formEl, data, loadErrors = []) {
             }
 
             if (form.encryptionKey) {
+                if (data.instanceAttachments) {
+                    fileManager.prefetchInstanceAttachments();
+                }
                 const saveDraftButton = document.querySelector(
                     '.form-footer#save-draft'
                 );
--- a/packages/enketo-express/public/js/src/module/file-manager.js
+++ b/packages/enketo-express/public/js/src/module/file-manager.js
@@ -15,6 +15,9 @@ const URL_RE = /[a-zA-Z0-9+-.]+?:\/\//;
 /** @type {Record<string, string>} */
 let instanceAttachments;
 
+/** @type {Map<string, Blob>} */
+const prefetchedBlobCache = new Map();
+
 /**
  * Initialize the file manager .
  *
@@ -41,6 +44,48 @@ function isWaitingForPermissions() {
  */
 function setInstanceAttachments(attachments) {
     instanceAttachments = attachments;
+    if (!attachments) {
+        prefetchedBlobCache.clear();
+    }
+}
+
+/**
+ * Pre-fetches all instance attachment URLs and caches the resulting Blobs.
+ * Should be called immediately after setInstanceAttachments while the
+ * server-side cache is still fresh.
+ *
+ * @return {Promise<void>}
+ */
+function prefetchInstanceAttachments() {
+    if (!instanceAttachments) {
+        return Promise.resolve();
+    }
+
+    const fetchPromises = Object.entries(instanceAttachments).map(
+        ([filename, url]) =>
+            fetch(url, { credentials: 'include' })
+                .then((response) => {
+                    if (!response.ok) {
+                        throw new Error(
+                            `Failed to fetch ${filename}: ${response.status}`
+                        );
+                    }
+
+                    return response.blob();
+                })
+                .then((blob) => {
+                    blob.name = filename;
+                    prefetchedBlobCache.set(filename, blob);
+                })
+                .catch((err) => {
+                    console.warn(
+                        `Failed to prefetch attachment "${filename}":`,
+                        err
+                    );
+                })
+    );
+
+    return Promise.all(fetchPromises).then(() => undefined);
 }
 /**
  * Obtains a url that can be used to show a preview of the file when used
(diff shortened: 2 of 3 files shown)


jnm:
Asked clarifying question at https://github.com/enketo/enketo/issues/1583#issuecomment-5454402252

jnm:
Hi @kelvin-muchiri, thanks for this contribution and answering my question on #1583.

I haven't evaluated this PR by hand yet, but I did ask the robot¹ to give it a review. At a glance, its gripes seem legitimate. Could you please address these or, if any are invalid, mention why? Thank you.

> Review of **enketo/enketo#1538** — `feat(encryption): support editing instances from encrypted forms` (touches `controller-webform.js`, `file-manager.js`, client tests). Six findings, most severe first:
> 
> **1. Cache key mismatch — the fix silently doesn't apply to filenames with special characters** (`file-manager.js:77`)
> `instanceAttachments` keys arrive from the server already escaped (`app/lib/media.js:229` → `escapeURLPath` + `encodeURIComponent` + `&amp;`/`&lt;` entities) — `getFileUrl` at lines 61–76 proves this by trying `encodeURIComponent(subject)` first. The new `prefetchedBlobCache` is keyed by that escaped name, but `getCurrentFiles` looks it up with `data-loaded-file-name`, which enketo-core sets to the **raw** instance value. Editing a submission with `my photo.jpg` → cache holds `my%20photo.jpg`, lookup misses, a string is pushed, and the save fails with the very `TypeError: Require Blob` this PR fixes. Key by the unescaped name, or check both variants like `getFileUrl` does. (Second-order: if an escaped name ever did match, `encryptor.js:213` would emit a `.enc` filename that doesn't match the submission XML, breaking decryption.)
> 
> **2. Prefetch is fire-and-forget** (`controller-webform.js:137`)
> No `await`/`return`, and it runs *after* `$('.main-loader').remove()`, so the form is interactive while blobs are still downloading. Encrypted submission + several multi-MB photos on mobile + quick edit-and-submit → empty/partial cache → same `Require Blob` failure. Await before revealing the form, or have the save path await the prefetch promise.
> 
> **3. Fetch failures swallowed into `console.warn`** (`file-manager.js:81`)
> Expired media-URL cache, 401/403, network blip — all silently ignored; the form looks fine and fails at submit with no indication of which attachment broke. For encrypted forms a missing blob is fatal, not cosmetic, so it should surface as a load error or retry at save time.
> 
> **4. Cache only invalidated on falsy attachments** (`file-manager.js:47`)
> `setInstanceAttachments(newMap)` without an intervening `null` call keeps all prior `filename → Blob` entries; loading a second instance in the same page context would encrypt and upload the *previous* instance's blob for a colliding name like `photo.jpg`. Also means tens of MB are retained for the page lifetime. Clearing unconditionally at the top costs nothing.
> 
> **5. Stale blob can be added alongside a newly selected file** (`file-manager.js:254`)
> `data-loaded-file-name` is removed asynchronously in the filepicker's `.finally` (enketo-core `filepicker.js:238-243`), so briefly both `input.files[0]` and the old attribute exist. Previously the extra entry was a harmless string; now it's a real Blob that gets encrypted and added to `<media>` and the element signature though the XML no longer references it. Narrow window, but it puts an unreferenced file into a signed manifest.
> 
> **6. Prefetched blobs skip the size check** (`file-manager.js:255`)
> They're pushed without `_processNameAndSize`/`isTooLarge`, so an existing attachment above `settings.maxSize` (e.g. uploaded via ODK Collect with a higher limit) is re-encrypted and re-uploaded, yielding a server-side rejection instead of the localized `filepicker.toolargeerror`.
> 
> #1 and #2 look blocking to me — both reproduce the exact bug the PR targets. #4 is a one-line fix worth taking now.

¹ Claude Opus 5 :robot:

kelvin-muchiri:
@jnm Thanks for the feedback. I have addressed each of the requested changes

PR #1541, opened 2026-06-22
DOMPurify: clear persistent config before sanitisation

alxndrsn:
By design, DOMPurify falls back to a global "persistent" configuration if set with `DOMPurify.setConfig(...)`.  This means that if `enketo-core` is included as a library in another package (e.g. `enketo/enketo-express`, `medic/cht-core` etc.), and the parent package calls `DOMPurify.setConfig()` anywhere, it will affect later behaviour of `DOMPurify.sanitize()` calls in `enketo-core`.

See:

* https://github.com/cure53/DOMPurify#persistent-configuration
* https://github.com/cure53/DOMPurify/issues/1134

### 🗒️ Checklist

- [ ] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`: no, i don't have access to private issue tracker
- [x] assign yourself
- [ ] I have verified this PR works by manually testing (see CONTRIBUTING.md):
    - [ ] Online form submission
    - [ ] Offline form submission
    - [ ] Saving offline drafts
    - [ ] Loading offline drafts
    - [ ] Editing submissions
    - [ ] Form preview
    - [x] None of the above

### 📣 Summary

Prevent DOMPurify config pollution.

### 💭 Notes

By design, DOMPurify falls back to a global "persistent" configuration if set with `DOMPurify.setConfig(...)`.  This means that if `enketo-core` is included as a library in another package (e.g. `enketo/enketo-express`, `medic/cht-core` etc.), and the parent package calls `DOMPurify.setConfig()` anywhere, it will affect later behaviour of `DOMPurify.sanitize()` calls in `enketo-core`.

See:

* https://github.com/cure53/DOMPurify#persistent-configuration
* https://github.com/cure53/DOMPurify/issues/1134

### 👀 Preview steps

Template:

2. do this
3. do that
4. 🔴 [on main] notice that this isn't anywhere BUT it should be here
5. 🟢 [on PR] notice that this is here

### DIFF
--- a/packages/enketo-core/src/js/sanitization-utils.js
+++ b/packages/enketo-core/src/js/sanitization-utils.js
@@ -464,6 +464,7 @@ function sanitizeSvg(svgElement) {
     DOMPurify.addHook('afterSanitizeAttributes', afterSanitizeAttributesHook);
 
     try {
+        DOMPurify.clearConfig();
         DOMPurify.sanitize(sanitizedSvg, DOMPURIFY_SVG_CONFIG);
     } finally {
         DOMPurify.removeHook('uponSanitizeElement', uponSanitizeElementHook);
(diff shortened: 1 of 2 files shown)


jamesrkiger (approved):
Thanks for noticing this and for submitting a fix, @alxndrsn.

PR #1544, opened 2026-06-22
DOMPurify: clear hooks before sanitisation

alxndrsn:
DOMPurify acts as a singleton, so "hooks" added for any use will persist, perhaps beyond the expectations of developers.

This means that if `enketo-core` is included as a library in another package (e.g.  `enketo/enketo-express`, `medic/cht-core` etc.), and the parent package calls `DOMPurify.addHook()` anywhere, it will affect later behaviour of `DOMPurify.sanitize()` calls in `enketo-core`.

Similar to https://github.com/enketo/enketo/pull/1541, this is a surprising feature of DOMPurify's API.

Ref:

* https://github.com/cure53/DOMPurify/issues/914
* https://github.com/cure53/DOMPurify/issues/1132
* https://github.com/cure53/DOMPurify/issues/1133

### DIFF
--- a/packages/enketo-core/src/js/sanitization-utils.js
+++ b/packages/enketo-core/src/js/sanitization-utils.js
@@ -460,11 +460,13 @@ function sanitizeSvg(svgElement) {
         }
     };
 
+    DOMPurify.clearConfig();
+    DOMPurify.removeAllHooks();
+
     DOMPurify.addHook('uponSanitizeElement', uponSanitizeElementHook);
     DOMPurify.addHook('afterSanitizeAttributes', afterSanitizeAttributesHook);
 
     try {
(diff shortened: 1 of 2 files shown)


packages/enketo-core/src/js/sanitization-utils.js
```diff
@@ -460,11 +460,13 @@ function sanitizeSvg(svgElement) {
         }
     };
 
+    DOMPurify.clearConfig();
+    DOMPurify.removeAllHooks();
```
alxndrsn, line 464: Seemed logical to group both "reset" commands.

jamesrkiger (approved):
This is a weird quirk of DOMPurify indeed. Thanks for highlighting it (again).

PR #1547, opened 2026-06-23
fix(draw): escape user-supplied attribute values DEV-2339

jamesrkiger:
### 📣 Summary

Fix XSS vulnerability in draw/annotate widget by escaping accept and capture attributes before HTML interpolation.

### 👀 Preview steps

Template:

1. Open a form with an annotate widget where mediatype is set to `image/*">`
2. On main, the js will be executed and the alert will fire
3. On this branch, the js will not be executed and the alert will not fire


pauloamorimbr (approved):
LGTM!

PR #1548, opened 2026-06-25
fix(autocomplete): script execution risk DEV-2341

pauloamorimbr:
### 📣 Summary

This PR fixes a security risk where the `listId` of an autocomplete field could be used to inject unwanted javascript code into the form.

### 💭 Notes
- Replaced `createContextualFragment` string interpolation with
`document.createElement` + `setAttribute`. The DOM API treats the value as a
literal string and never parses it as HTML, so no character in `listId` can
escape the attribute context.
- Added a test case to cover the fixed security violation
- I've investigated other solutions, but turned out this was the safest, since the `listId` is connected to other parts of Enketo that extract and save lists for later use.
- I've also checked all of the widgets and some other parts of code for similar vulnerabilities (unescaped attribute value set directly to the page)  and found none. 

### 👀 Preview steps
1. ℹ️ run Enketo locally
2. open for preview the test form on Enketo as in `http://localhost:8005/preview?form=http://localhost:3000/autocomplete_xss.xml` (verify form server port if using Enketo composer)
3. click the autocomplete input component
4. 🔴 [on main] javascript code executes and an alert appears
5. 🟢 [on PR] no code is executed.


packages/enketo-core/src/widget/select-autocomplete/autocomplete.js
```diff
@@ -70,13 +70,15 @@ class AutocompleteSelectpicker extends Widget {
...
-            .createRange()
-            .createContextualFragment(
-                `<input type="text" class="ignore widget autocomplete" list="${listId}" />`
-            );
+        const fakeInput = document.createElement('input');
+        fakeInput.type = 'text';
+        fakeInput.className = 'ignore widget autocomplete';
+        // Setting the attribute directly is safe because the browser will
+        // automatically escape any malicious characters in the listId and
+        // consider it a literal string, preventing XSS attacks.
```
jamesrkiger:
```suggestion
        // Setting the attribute directly is safe because it is being added as a literal string
```

packages/enketo-core/test/forms/autocomplete_xss.xml
jamesrkiger, line 1: I'd say let's not include this if we're not using it in a spec test.

packages/enketo-core/test/spec/widget.autocomplete.spec.js
```diff
@@ -2,3 +2,43 @@ import AutocompleteWidget from '../../src/widget/select-autocomplete/autocomplet
...
+        const control = fragment.querySelector('input[list]');
+        // Confirm the widget receives the decoded, raw malicious value.
+        if (control.getAttribute('list') !== maliciousListId) {
+            throw new Error(
+                'Test setup error: list attribute was not decoded correctly'
+            );
+        }
+
+        Promise.resolve()
+            .then(() => new AutocompleteWidget(control))
```
jamesrkiger: Do we need to wrap this in a promise?

jamesrkiger (approved):
Left a few minor comments. Good to merge after that.

PR #1554, opened 2026-07-02
fix(geopicker): input label element covering map on small screens DEV-2348

jamesrkiger:
### 📣 Summary

Fixed a bug that prevented interaction with the geopicker map when displaying RTL text on small screens.

### 👀 Preview steps

1. Create a project with this form:
[rtl-map.xlsx](https://github.com/user-attachments/files/29595575/rtl-map.xlsx)
2. Open the form in enketo on a small screen
3. On main, notice that the top coordinate input (below the map) label element covers the entire map, so you can't click on the map
4. On PR branch, verify that this is no longer the case.


pauloamorimbr (approved):
LGTM! 👌🏻

PR #1555, opened 2026-07-02
fix(audioWidget): detect invalid uploaded audio file DEV-2383

pauloamorimbr:
### 📣 Summary

Now files selectd for uploading are checked to be valid audio files

### 💭 Notes

Audio widget was accepting any file renamed to an audio file extension as uploaded. That was causing the file to not be parsed properly and player to not work, but without a check the state of the widget changed and a placeholder information was being displayed.
That was opening possibility to send invalid audio files to the server.

Now a validation function and tests were included to make sure the selected file is playable before accepting it.

### 👀 Preview steps

1. ℹ️ have an account and a project with an audio question
2. rename a not audio file to .mp3
3. Click "upload audio file" and pick the renamed file
4. 🔴 [on main] file is accepted, but cannot be played back. Information displays the 1:24 placeholder
5. 🟢 [on PR] a message is displayed to the user and the file is not accepted

### DIFF
--- a/packages/enketo-core/src/widget/audio/audio.js
+++ b/packages/enketo-core/src/widget/audio/audio.js
@@ -52,6 +52,17 @@ class AudioWidget extends Widget {
             // Handle file input change event for uploading audio files
             const file = event.target.files[0];
             if (file) {
+                const isValid = await this.isValidAudioFile(file);
+                if (!isValid) {
+                    this.element.type = 'text';
+                    this.showActionSelectStep();
+                    dialog.alert(
+                        t('audioRecording.error.invalidAudioFile.msg'),
+                        t('audioRecording.error.invalidAudioFile.heading'),
+                        'normal'
+                    );
+                    return;
+                }
                 await this.updateValue(file); // Update the widget with the uploaded file
                 this.fileName = this.postFixFilename(file.name); // Update the filename with a unique timestamp postfix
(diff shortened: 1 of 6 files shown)


packages/enketo-core/src/widget/audio/audio.js
```diff
@@ -489,6 +500,39 @@ class AudioWidget extends Widget {
...
+     * @param {File} file - The file to verify.
+     * @returns {Promise<boolean>} - Resolves to true if the file is playable audio, false otherwise.
+     */
+    isValidAudioFile(file) {
+        return new Promise((resolve) => {
+            const url = URL.createObjectURL(file);
+            const audio = new Audio();
+            const cleanup = () => URL.revokeObjectURL(url);
+            audio.addEventListener(
+                'canplay',
```
jamesrkiger, line 517: Have you considered using `loadedmetadata` here? I'm wondering if that might be good enough for what we're trying to check.

jamesrkiger (approved):
Left one comment, but otherwise lgtm

packages/enketo-core/src/widget/audio/audio.js
pauloamorimbr: I've seen issues with metadata before. 
Here in the widget we added an external library to calculate and fill in the duration metadata to recorded audio, so I know some audio formats may lack metadata and maybe they won't fire the event. For that I considered `canplay` to be safer in the context.

PR #1571, opened 2026-07-15
feat(filepicker): add support for HEIC/HEIF file extensions DEV-2396

pauloamorimbr:
### 📣 Summary

This PR adds support for listing .heic and .heif files in the file picker dialog for images.

### 💭 Notes

These file formats are becoming very popular due to iPhones, and KPI was changes to support them already. It's already possible for the user to pick those files, but they need to change the "File type" in the file browser dialog.
Now the .heic and .heif files are listed when "Image Files" is selected.

Before:
image

After:
image

### 👀 Preview steps
1. Have .heic and .heif files available
1. ℹ️ open a form with a image picker
2. click to select image
3. 🔴 [on main] notice that .heic and .heif files are not listed
4. 🟢 [on PR] notice that .heic and .heif files are listed by default

### DIFF
--- a/packages/enketo-core/src/widget/draw/draw-widget.js
+++ b/packages/enketo-core/src/widget/draw/draw-widget.js
@@ -615,7 +615,11 @@ class DrawWidget extends Widget {
                       'darkgray',
                   ];
         props.touch = support.touch;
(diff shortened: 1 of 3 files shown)


jamesrkiger (approved):
LGTM

Issue #243, opened 2018-09-10
Referring to previous repeat: calculations not triggered


enketo-issue-mover:
_From @MartijnR on August 3, 2015 22:11_

https://docs.google.com/spreadsheets/d/13tMk8fNhTdAaarBxHotAaKLT7dqw-60gODjiExOlgts/edit?usp=sharing

_Copied from original issue: kobotoolbox/enketo-express#278_

PR #1572, opened 2026-07-16
chore(security): Bump library versions DEV-2423

pauloamorimbr:
Triages and resolves the Dependabot alerts open on this repo (github.com/enketo/enketo/security/dependabot), fixing them directly via version bumps and/or targeted yarn `resolutions` overrides for transitive dependencies, verified with full local test runs and a Docker-based CI reproduction rather than relying on Dependabot's automatic PRs.

### Direct dependency bumps

| Package | Before | After |
|---|---|---|
| `vitest` (+ `@vitest/coverage-istanbul`, `@vitest/runner`, `@vitest/utils`) | `^0.34.6` / `^0.34.3` | `^4.1.10` |
| `vite` | `^4.5.3` | `^6.4.3` |
| `vite-node` | `^2.1.1` | `^3.2.4` |
| `undici` (enketo-transformer) | `^5.28.4` | `^6.24.0` |
| `markdown-it` (enketo-express) | `^13.0.2` | `^14.3.0` | 
| `node-sass` → `sass` | `^9.0.0` | `sass@1.99.0` (exact) | 

`sass` is pinned to an **exact** `1.99.0` rather than a caret range: `1.100.0`+ raises the `engines.node` requirement to `>=20.19.0`, which would break the Node `20.18.1` leg of the CI matrix. `1.99.0` only requires `>=14.0.0` and is otherwise fully current.

Similarly, `vite` is capped at `^6.4.3` rather than the latest `7.x`: `vite@7` requires Node `^20.19.0 || >=22.12.0`, again breaking the `20.18.1` CI leg. `6.4.3` still contains the fixes for all the alerts above.

### Transitive fixes via yarn `resolutions`

Yarn classic doesn't let Dependabot bump these directly since they're several levels deep. Added to the root `package.json`:

```jsonc
"resolutions": {
    "markdown-eslint-parser/eslint/file-entry-cache/flat-cache/flatted": "^3.4.2", // flatted prototype pollution
    "**/globule/minimatch": "^3.1.5",
    "mocha/minimatch": "^4.2.5",                      // ReDoS in minimatch
    "mocha/serialize-javascript": "^7.0.3",           // #105/#242 XSS / CPU exhaustion DoS
    "markdown-eslint-parser/eslint/inquirer/external-editor/tmp": "^0.2.7", // #127 tmp symlink write
    "markdown-it": "^14.3.0",                          // covers the yarn.lock-only markdown-it alerts too
    "@xmldom/xmldom": "^0.9.8",
    "vite": "^6.4.3",
    "cookie": "^0.7.2",                                // #95 cookie out-of-bounds chars
    "qs": "^6.15.3",                                    // #244 qs.stringify DoS
    "mocha/js-yaml": "^4.3.0",                          // #141 prototype pollution, #277/#278 quadratic DoS
    "grunt/js-yaml": "^3.15.0"
}
```

Note: Yarn classic's `"parent/child": "version"` resolution syntax only takes effect when `parent` is itself a **root-level** declared dependency (e.g. `mocha`, `grunt`). For packages nested deeper in the tree whose direct parent isn't top-level (e.g. `csurf`, `jstransformer-markdown-it`), a bare global override (`"cookie": "^0.7.2"`) was used instead, since the scoped syntax silently no-ops in that case.

### Deferred

- **`uuid` — Missing buffer bounds check in v3/v5/v6 when `buf` is provided (#243)**: `uuid` is purely transitive (no direct dependent declares it at the root), currently resolves to `8.3.2`. Left unaddressed in this PR — flagging for follow-up rather than forcing an override without being able to verify no other transitive consumer breaks on the newer major.

### Extra changes required as a consequence of the bumps

- **`packages/enketo-core/Gruntfile.js`, `packages/enketo-express/Gruntfile.js`**: `require('node-sass')` → `require('sass')`, and `implementation: nodeSass` → `implementation: sass` in the `sass` Grunt task options.
- **`packages/enketo-transformer/vite.config.ts`** (Vitest 0.34 → 4 is a major jump with several breaking changes):
  - `UserConfig` is no longer exported from `vitest/config`; now imported from `vite` directly (module-augmented in v4).
  - `build.target: 'modules'` was a special Vite value that's been removed; replaced with its closest equivalent, `'esnext'`.
  - `test.threads`/`test.singleThread` were replaced by `test.maxWorkers: 1` + `test.isolate: false` (Vitest 4's pool rework) — still needed because the libxmljs/libxslt native N-API bindings aren't thread-safe.
  - `resolve.alias` replacement paths changed from relative strings to absolute (`resolve(__dirname, './src/dom/web/index.ts')`); relative paths silently fail to resolve under Vitest 4's nested dev server used by the "web" test target.
  - The `/@fs/` alias regex was fixed from `/^\/@fs\/(.*)/ ` to `/^\/@fs(\/.*)/` — the old pattern consumed the path's leading slash, producing a bare-specifier-looking string (`home/...`) that Node's ESM resolver misread as a package name (`Cannot find package 'home'`).
  - `fileName(format, entryName)` needed explicit param types (inference regressed under the new `vite`/`vitest` type versions).
- **`packages/enketo-core/test/spec/widget.audio.spec.js`** + new **`packages/enketo-core/test/helpers/wait-for.js`**: Not directly related, but surfaced by the vitest version bump — `AudioWidget`'s upload tests relied on a fixed `setTimeout(fn, 0)` to synchronize with an `async` change handler that awaits a stubbed promise and then a real `FileReader` read. This is not a reliable barrier (`FileReader` completion isn't tied to a fixed macrotask delay), and became flaky in CI. Replaced with a small shared `waitFor(predicate)` polling helper and converted the affected tests to `async`/`await` (matching the async-test convention already used elsewhere in this suite, e.g. `form.spec.js`).


packages/enketo-core/test/spec/widget.audio.spec.js
```diff
@@ -223,20 +224,17 @@ describe('AudioWidget', () => {
...
-                expect(postFixSpy.calledOnce).to.equal(true);
-                expect(postFixSpy.firstCall.args[0]).to.equal(
-                    'same-name-test.webm'
-                );
-                expect(widget.fileName).to.match(
-                    /^same-name-test-\d{8}_\d{6}\.webm$/
-                );
-                done();
-            }, 0);
+            await waitFor(() => postFixSpy.calledOnce);
```
pauloamorimbr, line 227: This was needed because setTimeout started failing on new version.

jamesrkiger (changes requested):
Generally looks good, but I think the Sass deprecation noise on build (measured in thousands of lines of output) is bad enough to warrant silencing the deprecation warnings before merging this.

jamesrkiger (approved):
lgtm

PR #1576, opened 2026-07-29
feat(store): implement IndexedDB schema and reconnection logic DEV-2436

pauloamorimbr:
### 📣 Summary

Now a IDBDatabase connection that gets closed is automatically restablished

### 💭 Notes
On iOS, when memory is under pressure (e.g. after switching to several memory-heavy apps), Safari can forcibly close a page's IndexedDB connection in the background, even if the Enketo tab stays open. Previously, once this happened, the app kept using the now-dead connection and any attempt to save a draft or submit a form would fail with an error like `Failed to execute 'transaction' on 'IDBDatabase'`, putting users at risk of losing their work. This issue is reported to happen on iOS in situations of high memory usage; it's a rare bug that's hard to reproduce and verify on demand, since it depends on iOS's own memory-reclaim behavior rather than anything within Enketo's control.

This PR detects when the underlying connection is closed unexpectedly and automatically reopens it, so that saving and submitting continue to work without requiring the user to reload the page.

- Added unit tests in [store.spec.js](packages/enketo-express/test/client/store.spec.js) covering the reconnect flow.
- I decided to leave the `console.error` and `console.log` since this is not a common issue and it helps to map an invisible behavior if needed

### 👀 Preview steps

1. ℹ️ Open a form in Enketo and start filling it out (online or offline).
2. Open DevTools → Application → IndexedDB → delete/clear the `enketo` database (simulating iOS force-closing the connection).
3. 🔴 [on main] Continue filling the form and submit/save a draft — this fails with an `IDBDatabase` transaction error and the record may be lost.
4. 🟢 [on PR] Continue filling the form and submit/save a draft — a `IndexedDB connection was closed unexpectedly. Attempting to reconnect...` message appears in the console, followed by `IndexedDB connection was successfully recreated.`, and the save/submit succeeds normally.

### DIFF
--- a/packages/enketo-express/public/js/src/module/records-queue.js
+++ b/packages/enketo-express/public/js/src/module/records-queue.js
@@ -106,7 +106,9 @@ function save(action, record) {
             return result;
         })
         .then(({ enketoId }) => formCache.get({ enketoId }))
-        .then((survey) => setLastSavedRecord(survey, record))
+        .then((survey) =>
+            survey ? setLastSavedRecord(survey, record) : undefined
+        )
         .then(_updateRecordList)
         .then(() => result);
 }
--- a/packages/enketo-express/public/js/src/module/store.js
+++ b/packages/enketo-express/public/js/src/module/store.js
@@ -35,6 +35,180 @@ const version = 4;
 
 const REMOVE_RECORD_NAME_UNIQUENESS_VERSION = 4;
 
+// the enketo db.js schema, shared between the initial connection and any
+// reconnection attempts (see `_openDatabase`)
+const schema = {
+    // the surveys
+    surveys: {
+        key: {
+            keyPath: 'enketoId',
+            autoIncrement: false,
+        },
+        indexes: {
+            enketoId: {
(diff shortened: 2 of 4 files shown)


jamesrkiger:
@pauloamorimbr I'm running into this during testing your repro steps. The db connection gets restored successfully but I still can't submit:

image

Looks like it might be a downstream issue we only see now because of your fix. But you didn't mention it, so I'm wondering if I'm testing correctly.

pauloamorimbr:
> @pauloamorimbr I'm running into this during testing your repro steps. The db connection gets restored successfully but I still can't submit:
> 
> image
> Looks like it might be a downstream issue we only see now because of your fix. But you didn't mention it, so I'm wondering if I'm testing correctly.

Good catch. 
This is an issue related to IDB being not present for some reason, but not directly related to the original issue.
Fixed the error by testing and bypassing the function call.
This is a optional caching being done, so even in the case of this error there's no issue in skipping the function.

jamesrkiger (approved):
LGTM

Issue #1574, opened 2026-07-23
Google Maps tiles don't work with EE 7.6.2+


lognaturel:
**Describe the bug**
When EE 7.6.2+ is configured to use Google Maps, there's console error `TypeError: L.GridLayer.GoogleMutant is not a constructor`

**To Reproduce**
1. Use EE 7.6.2+
2. Configure Google Maps
3. Try to load a form with a map-based question type

**Expected behavior**
Google Maps works

**Screenshots**

**Browser and OS (please complete the following information):**

**Additional context**
This is related to the upgrade of [GoogleMutant](https://gitlab.com/IvanSanchez/Leaflet.GridLayer.GoogleMutant) that happened in 7.6.2. It's not exactly clear what the problem is but the latest version of `GoogleMutant` doesn't provide any benefit to Enketo so downgrading is an easy fix.

PR #1568, opened 2026-07-15
gha/dependabot-automerge: fix `if` syntax

alxndrsn:
I assume this is what was intended.

Closes #1545

### 🗒️ Checklist

- [ ] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
- [ ] assign yourself
- [ ] fill in the template below and delete template comments
- [ ] update all related docs (README, inline, etc.), if any
- [ ] I have verified this PR works by manually testing (see CONTRIBUTING.md):
    - [ ] Online form submission
    - [ ] Offline form submission
    - [ ] Saving offline drafts
    - [ ] Loading offline drafts
    - [ ] Editing submissions
    - [ ] Form preview
    - [ ] None of the above
- [x] review thyself: read the diff and repro the preview as written
- [x] undraft PR & confirm that CI passes
- [ ] request reviewers & improve according to review
- [ ] delete this section before merging

### 👀 Preview steps

Template:

1. ℹ️ have an account and a project
2. do this
3. do that
4. 🔴 [on main] notice that this isn't anywhere BUT it should be here
5. 🟢 [on PR] notice that this is here

PR #1567, opened 2026-07-15
enketo-express: disable X-Powered-By header

alxndrsn:
Closes #1559

### 🗒️ Checklist

- [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
- [x] assign yourself
- [x] fill in the template below and delete template comments
- [x] update all related docs (README, inline, etc.), if any
- [x] I have verified this PR works by manually testing (see CONTRIBUTING.md):
    - [ ] Online form submission
    - [ ] Offline form submission
    - [ ] Saving offline drafts
    - [ ] Loading offline drafts
    - [ ] Editing submissions
    - [ ] Form preview
    - [x] None of the above
- [x] review thyself: read the diff and repro the preview as written
- [x] undraft PR & confirm that CI passes
- [ ] request reviewers & improve according to review
- [ ] delete this section before merging

PR #1566, opened 2026-07-15
enketo-express: update OpenSans fonts

alxndrsn:
* fixes bugs with current OpenSans fonts
* makes font files smaller

Copied existing font files from `packages/enketo-transformer-web`.

These can be compared:

```sh
diff examples/enketo-transformer-web/public/OpenSans-Bold-webfont.woff packages/enketo-express/public/fonts/OpenSans-Bold-webfont.woff
diff examples/enketo-transformer-web/public/OpenSans-Regular-webfont.woff packages/enketo-express/public/fonts/OpenSans-Regular-webfont.woff
```

Testing:

1. fonts were compared visually by uploading to https://fontdrop.info/ and comparing every glyph
2. when uploading to https://fontdrop.info/, glyph errors can be seen in firefox console for old `.woff` files, but absent for updated `.woff` files

Closes #1540

### 🗒️ Checklist

- [x] draft PR with a title `<type>(<scope>)<!>: <title> DEV-1234`
- [x] assign yourself
- [x] fill in the template below and delete template comments
- [x] update all related docs (README, inline, etc.), if any
- [ ] I have verified this PR works by manually testing (see CONTRIBUTING.md):
    - [ ] Online form submission
    - [ ] Offline form submission
    - [ ] Saving offline drafts
    - [ ] Loading offline drafts
    - [ ] Editing submissions
    - [ ] Form preview
    - [x] None of the above
- [x] review thyself: read the diff and repro the preview as written
- [x] undraft PR & confirm that CI passes
- [ ] request reviewers & improve according to review
- [ ] delete this section before merging

PR #1586, opened 2026-08-10
test(coverage): add regression tests for #1564, #1565, #1567 DEV-2654

pauloamorimbr:
### 💭 Notes

Adds automated test coverage for a few recently-merged fixes that previously had none, closing gaps identified while reviewing #1564, #1565, #1566, #1567 and #1568.

**What's covered:**

- `packages/enketo-core/test/spec/widget.geo.spec.js`
  - Regression test for #1564 / #1574 - asserts `L.gridLayer.googleMutant` registers as a working constructor (would have failed with `leaflet.gridlayer.googlemutant@0.16.0` + `leaflet@1.9.4`, the exact combination that broke Google Maps tiles).
  - Coverage for #1565 - asserts a configured `map.referrerPolicy` actually reaches the Leaflet `TileLayer` instance via `_getLeafletTileLayer`/`_getTileOptions`, plus a default-value check.
- `packages/enketo-express/test/server/express-config.spec.js`
  - Coverage for #1567 - asserts the Express app does not send an `X-Powered-By` header.

All three tests were verified to fail against the pre-fix code and pass now that the corresponding PRs merged into `main`.

No behavior changes - test-only PR. #1566 (font update) and #1568 (GH Actions `if` syntax) aren't unit-testable in this suite, so no coverage was added for those.


packages/enketo-express/test/server/express-config.spec.js
jamesrkiger, line 1: I'm a little skeptical of the value of this test. Is there any scenario in which it would start failing other than someone deleting the line that was added in the relevant PR?

jamesrkiger (approved):
Left one comment about a test that I think might be worth dropping, but otherwise LGTM

packages/enketo-express/test/server/express-config.spec.js
pauloamorimbr: Yeah...I think was a bit overboard thinking again about it. I'll remove that one. Thank you.

PR #1592, opened 2026-08-11
fix(filepicker): image previews for photo widget DEV-2658

pauloamorimbr:
### 📣 Summary

This change restores image previews for photo/file widgets by keeping the original media type stable even after the widget expands the `accept` attribute for HEIC/HEIF support.

### 📖 Description

The preview logic in the file picker widget was relying on the `accept` attribute value after it had been expanded for better OS file-picker compatibility. That caused image widgets to stop rendering a preview because the preview code no longer matched the expected media type.

This PR captures the original media type up front, keeps it available on the widget props for compatibility, and uses that stable value for preview and resize decisions.

### 💭 Notes

- The underlying fix is localized to the file picker widget.
- Preview behavior for image widgets should now be restored without affecting the HEIC/HEIF file-picker compatibility change.
- The audio preview capabilities were removed since they are not being used anymore since the audio recording widget was implemented

### 👀 Preview steps

2. open a form that includes an image/photo upload widget
3. select or load an image file
4. 🟢 notice that the image preview is shown again in the widget

### DIFF
--- a/packages/enketo-core/src/widget/file/filepicker.js
+++ b/packages/enketo-core/src/widget/file/filepicker.js
@@ -32,11 +32,21 @@ class Filepicker extends Widget {
         );
         const that = this;
 
+        // Capture the original media type before the accept attribute is
+        // potentially expanded below, and resolve it to flags up front. The
+        // `accept` attribute itself is not a reliable source of truth after
+        // that point (see below), so nothing else in this widget should read
+        // or compare against it directly.
+        this.mediaType = this.element.getAttribute('accept');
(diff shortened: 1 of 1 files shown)


jamesrkiger (approved):
lgtm

tiritea:
FWIW there's a few folk anxiously hanging out for this fix, so probably the sooner we can get it deployed the better. Thnx.

tiritea:
@pauloamorimbr  Thanks much for the quick turnaround on this!

Issue #1590, opened 2026-08-10
Consider encrypting JWTs instead of signing them


alxndrsn:
Highlighted in https://github.com/enketo/enketo/pull/1587

JWTs generated at: https://github.com/enketo/enketo/blob/eef1e2560a0b574f17ea865264d4080a9342e689/packages/enketo-express/app/controllers/authentication-controller.js#L97-L103

These are signed JWTs, rather than encrypted JWEs, so user's password is included in plaintext.  These could instead be encrypted, which would mean a leaked token would not expose the password.  Currently there is no token expiration (see #1589), so a leaked encrypted token would continue to be valid until the encryption key is rotated.

Also note that the current JWT lib (`jwt-simple`) does not support encryption, so a different lib would be required.

---

Related: https://github.com/enketo/enketo/issues/1589

Issue #1589, opened 2026-08-10
Consider setting (& checking) expiration field for JWTs


alxndrsn:
Highlighted in https://github.com/enketo/enketo/pull/1587

There's no expiration set when tokens are generated: https://github.com/enketo/enketo/blob/eef1e2560a0b574f17ea865264d4080a9342e689/packages/enketo-express/app/controllers/authentication-controller.js#L97-L103

Would presumably need some mechanism to refresh tokens.

---

Related: https://github.com/enketo/enketo/issues/1590

PR #1594, opened 2026-08-13
enketo-express: add exp/iat claims to JWTs and handle expiry on decode

anupamme:
## Summary

Addresses #1589.

JWTs generated at login had no expiration claim, meaning a leaked or stolen token was valid indefinitely. This PR:

- Adds `iat` and `exp` claims to the JWT payload in `setToken()`:
  - **"Remember me"** sessions: `exp` = 30 days (matching the existing cookie `maxAge`)
  - **Session-only** logins: `exp` = 24 hours
- Wraps `jwt.decode()` in `getCredentials()` with a try/catch so that an expired token (or any other decode error) is treated as absent credentials rather than propagating an unhandled exception. `jwt-simple` already enforces the `exp` claim on decode.
- Adds `test/server/user-model.spec.js` covering:
  - Valid non-expired token → credentials returned
  - Expired token → `null` (no throw)
  - Legacy token without `exp` → still decoded (backward-compatible)
  - Missing cookie → `null`

**Out of scope:** Token refresh and JWT encryption (#1590) are left as separate follow-ups per the issue discussion.

## Test plan

- [ ] Run `npm run test-server` in `packages/enketo-express` — all tests pass including the new `user-model.spec.js`
- [ ] Log in with "remember me" → decode the cookie JWT → confirm `exp` ≈ now + 30 days
- [ ] Log in without "remember me" → confirm `exp` ≈ now + 24 hours
- [ ] Manually set a cookie with an expired JWT → confirm the request is treated as unauthenticated (401/redirect to login), not a 500

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/enketo-express/app/controllers/authentication-controller.js
+++ b/packages/enketo-express/app/controllers/authentication-controller.js
@@ -94,10 +94,17 @@ function setToken(req, res) {
     const maxAge = 30 * 24 * 60 * 60 * 1000;
     const returnUrl = req.query.return_url || '';
 
+    const nowSecs = Math.floor(Date.now() / 1000);
+    const expSecs = req.body.remember
+        ? nowSecs + 30 * 24 * 60 * 60
+        : nowSecs + 24 * 60 * 60;
+
     const token = jwt.encode(
         {
             user: username,
             pass: req.body.password,
+            iat: nowSecs,
+            exp: expSecs,
         },
         req.app.get('encryption key')
     );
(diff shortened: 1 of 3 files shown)


alxndrsn:
> **Out of scope:** Token refresh and JWT encryption (#1590) are left as separate follow-ups per the issue discussion.

I suspect discussion of token refresh would need to be resolved before introducing token expiration.

PR #1595, opened 2026-08-13
enketo-express: encrypt auth JWTs (JWS → JWE)

### LINKED ISSUE #1590: Consider encrypting JWTs instead of signing them

anupamme:
## Summary

Closes #1590.

The auth cookie stored a **signed** JWT (`jwt-simple`, HS256) with `{ user, pass }` in the payload. Because signing does not encrypt, anyone who reads the cookie can base64-decode the payload and recover the plaintext password without any key material.

This PR switches to **encrypted JWTs (JWE)** using [`jose`](https://github.com/panva/jose) with `alg: dir, enc: A256GCM`:

- **`app/lib/encryption.js`** (new) — derives a 32-byte AES key from the existing `encryption key` config string via SHA-256, so no config changes are required
- **`authentication-controller.js`** — `setToken` uses `EncryptJWT`; also adds `iat`/`exp` claims (24 h default, 30 d with "remember me") so tokens expire server-side
- **`user-model.js`** — `getCredentials` uses `jwtDecrypt`; any decryption failure (expired, tampered, or pre-migration signed token) returns `null`, forcing re-login
- **`package.json`** — replaces `jwt-simple` with `jose ^5.9.6`
- All `getCredentials` / `getHostURLOptions` call sites updated for the new async API
- **`test/server/user-model.spec.js`** (new) — 5 unit tests: valid JWE, expired token, old JWS token (migration compat), missing cookie, token auth type

**Breaking change for existing sessions:** tokens issued before this change are signed JWTs (3-part compact JWS). `jwtDecrypt` rejects them as invalid JWE, returning `null` — users will need to log in again once after deployment.

## Test plan

- [ ] `node_modules/.bin/mocha packages/enketo-express/test/server/user-model.spec.js` — all 5 tests pass
- [ ] Log in via `/login`, inspect the `__enketo_` cookie — should be a 5-part compact JWE string instead of a 3-part JWS
- [ ] Submit a form with basic auth — confirm credentials are forwarded correctly to the OpenRosa server
- [ ] Log out and log back in with "remember me" checked — confirm 30-day expiry token is issued
- [ ] Paste an old-style signed token into the cookie manually — confirm the app treats it as unauthenticated (redirects to login)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/enketo-express/app/controllers/authentication-controller.js
+++ b/packages/enketo-express/app/controllers/authentication-controller.js
@@ -5,7 +5,8 @@
 const csrfProtection = require('csurf')({
     cookie: true,
 });
-const jwt = require('jwt-simple');
+const { EncryptJWT } = require('jose');
+const { deriveEncryptionKey } = require('../lib/encryption');
 const express = require('express');
 
 const router = express.Router();
@@ -88,19 +89,28 @@ function logout(req, res) {
 /**
  * @param {module:api-controller~ExpressRequest} req - HTTP request
  * @param {module:api-controller~ExpressResponse} res - HTTP response
+ * @param {Function} next - Express callback
  */
-function setToken(req, res) {
+async function setToken(req, res, next) {
     const username = req.body.username.trim();
     const maxAge = 30 * 24 * 60 * 60 * 1000;
     const returnUrl = req.query.return_url || '';
 
-    const token = jwt.encode(
-        {
-            user: username,
-            pass: req.body.password,
-        },
-        req.app.get('encryption key')
-    );
+    let token;
+    try {
+        const derivedKey = deriveEncryptionKey(req.app.get('encryption key'));
+        const nowSecs = Math.floor(Date.now() / 1000);
+        const expSecs = req.body.remember
+            ? nowSecs + 30 * 24 * 60 * 60
+            : nowSecs + 24 * 60 * 60;
+        token = await new EncryptJWT({ user: username, pass: req.body.password })
+            .setProtectedHeader({ alg: 'dir', enc: 'A256GCM' })
+            .setIssuedAt()
+            .setExpirationTime(expSecs)
+            .encrypt(derivedKey);
+    } catch (err) {
+        return next(err);
+    }
 
     // Do not allow authentication cookies to be saved if enketo runs on http, unless 'allow insecure transport' is set to true
     // This is double because the check in login() already ensures the login screen isn't even shown.
--- a/packages/enketo-express/app/models/user-model.js
+++ b/packages/enketo-express/app/models/user-model.js
@@ -2,7 +2,8 @@
  * @module user-model
  */
 
-const jwt = require('jwt-simple');
+const { jwtDecrypt } = require('jose');
+const { deriveEncryptionKey } = require('../lib/encryption');
 const url = require('url');
 // var debug = require( 'debug' )( 'user-model' );
 
@@ -12,9 +13,9 @@ const url = require('url');
  *
  * @static
  * @param {module:api-controller~ExpressRequest} req - HTTP request
- * @return {object|null} Credentials
+ * @return {Promise<object|null>} Credentials
  */
-function getCredentials(req) {
+async function getCredentials(req) {
     const auth = req.app.get('linked form and data server').authentication;
     const authType = auth.type.toLowerCase();
     let creds = null;
@@ -22,9 +23,17 @@ function getCredentials(req) {
     if (authType === 'basic') {
         const jwToken =
             req.signedCookies[req.app.get('authentication cookie name')];
-        creds = jwToken
-            ? jwt.decode(jwToken, req.app.get('encryption key'))
-            : null;
+        if (jwToken) {
+            try {
+                const derivedKey = deriveEncryptionKey(
+                    req.app.get('encryption key')
+                );
+                const { payload } = await jwtDecrypt(jwToken, derivedKey);
+                creds = { user: payload.user, pass: payload.pass };
+            } catch {
+                creds = null;
+            }
+        }
     } else if (authType === 'token') {
         const paramName = auth['query parameter'];
         if (!paramName) {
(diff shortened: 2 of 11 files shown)


packages/enketo-express/app/controllers/authentication-controller.js
```diff
@@ -88,19 +89,28 @@ function logout(req, res) {
 /**
  * @param {module:api-controller~ExpressRequest} req - HTTP request
  * @param {module:api-controller~ExpressResponse} res - HTTP response
+ * @param {Function} next - Express callback
  */
-function setToken(req, res) {
+async function setToken(req, res, next) {
     const username = req.body.username.trim();
```
jamesrkiger: Now that this is an async function (and because Enketo is still on Express 4), I think we need to be more careful with error handling since only sync throws get caught automatically. If username is undefined here the request will hang instead of failing now.

anupamme: addressed

jamesrkiger (approved):
Looks good to me. Thanks a lot for the contribution, @anupamme.

packages/enketo-express/app/controllers/authentication-controller.js
```diff
@@ -88,19 +89,29 @@ function logout(req, res) {
...
-    );
+    let username;
+    let token;
+    try {
+        username = req.body.username.trim();
+        const derivedKey = deriveEncryptionKey(req.app.get('encryption key'));
+        const nowSecs = Math.floor(Date.now() / 1000);
+        const expSecs = req.body.remember
+            ? nowSecs + 30 * 24 * 60 * 60
+            : nowSecs + 24 * 60 * 60;
```
jnm: Since the token now expires after 24h if the "Remember on this computer" option is not set, regardless of how long the session lasts, is there any reason to keep the cookie around (`maxAge`) after the token expires?

packages/enketo-express/app/models/user-model.js
```diff
@@ -12,19 +13,27 @@ const url = require('url');
...
-            : null;
+        if (jwToken) {
+            try {
+                const derivedKey = deriveEncryptionKey(
+                    req.app.get('encryption key')
+                );
+                const { payload } = await jwtDecrypt(jwToken, derivedKey);
+                creds = { user: payload.user, pass: payload.pass };
+            } catch {
+                creds = null;
```
jnm: This doesn't seem to do anything since `creds` was already initialized to `null`. How about being sure the `catch` doesn't swallow unrelated exceptions and logging the failure? Something like:
```js
if (!(err instanceof errors.JOSEError)) throw err;
debug('auth cookie rejected: %s', err.code);
```
If that makes sense, you'd also need to replace the commented-out `// var debug = require…` with `const debug = require('debug')('enketo:user-model');` (note the addition of the `enketo:` prefix). I think this was commented out just to satisfy the linter, since `debug()` was previously unused here, and it just never got updated with the prefix.

packages/enketo-express/app/models/user-model.js
```diff
@@ -12,19 +13,27 @@ const url = require('url');
...
             req.signedCookies[req.app.get('authentication cookie name')];
-        creds = jwToken
-            ? jwt.decode(jwToken, req.app.get('encryption key'))
-            : null;
+        if (jwToken) {
+            try {
+                const derivedKey = deriveEncryptionKey(
+                    req.app.get('encryption key')
+                );
+                const { payload } = await jwtDecrypt(jwToken, derivedKey);
```
jnm, line 31: I'd argue only this line should be within the `try`. If retrieving or deriving from `encryption key` fails, that's sysadmin or programmer error. And subsequently building the simple `creds` object should never fail

jnm (changes requested):
Thanks for the contribution. Would you be willing also to update `yarn.lock` and address the linter feedback so that the CI can pass? 🙏

packages/enketo-express/app/controllers/authentication-controller.js
anupamme: Addressed.

packages/enketo-express/app/models/user-model.js
anupamme: Addressed.

packages/enketo-express/app/models/user-model.js
anupamme: Addressed.

alxndrsn:
> **Breaking change for existing sessions:** tokens issued before this change are signed JWTs (3-part compact JWS). `jwtDecrypt` rejects them as invalid JWE, returning `null` — users will need to log in again once after deployment.

Is there also a breaking change here for new session lifetimes?  It seems like previous session lifetime was infinite, and now it will default to 24 hours.  Have I understood correctly?

anupamme:
Yes, you've understood correctly; worth calling out explicitly. Before this PR, jwt-simple's jwt.encode() was called with no exp claim, so the token itself never expired; the only thing bounding a session was the cookie's maxAge, and without "Remember on this computer" that was a plain session cookie (no maxAge at all), so in practice a session could last indefinitely; and even if the cookie expired client-side, the server would still honour a replayed old token forever, since nothing checked its age.
  
This PR gives the JWE a real exp claim (24h by default, 30 days with "remember"), enforced server-side by jwtDecrypt, with the cookie maxAge aligned to match. So yes, default sessions are now capped at 24h where they previously had no server-enforced bound.
  
This is intentional hardening rather than a regression, but happy to make the default configurable or pick a longer default. If 24h is too aggressive for existing deployments, let me know your preference.

alxndrsn:
> Yes, you've understood correctly; worth calling out explicitly.

Thanks!  I'm happy with a strong default - I just wanted to be clear on what might change for existing clients and integrations.

PR #1597, opened 2026-08-18
feat(errors): improve error messages DEV-243

pauloamorimbr:
### 📣 Summary

Error messages shown when a form can't be found no longer leak internal details (form IDs, server request URLs) and are clearer for end-users.

### 📖 Description

- The "form not found" error no longer includes the form ID.
- Failed requests to the OpenRosa server no longer include the internal server URL in the error message shown to users.
- For Spanish, French and Portuguese the translation of the messages were added, as they were part of the original PR. Other languages were removed to use `en` as the fallback until translations are added.

### 👀 Preview steps

1. ℹ️ Point Enketo at an inexistent form, e.g: http://ee.kobo.local/ABCDEFGH
2. 🔴 [on main] Notice that the error message says "Survey with this ID not found"
3. 🟢 [on PR] Notice that the error message says "No form found with this URL, please double check."
1. ℹ️ Create a form, open in Enketo, Archive the form and reload Enketo
2. 🔴 [on main] Notice that the error message includes a long explanation and the form ID
3. 🟢 [on PR] Notice the error message is now generic ("This form is no longer available.")

The other error covered by the PR isn't really testable, being a backend call, but the situation is covered in the test added to the PR.


jamesrkiger (approved):
LGTM

PR #1600, opened 2026-08-24
fix(submissions): keep defaults for multiple submissions DEV-2716

jamesrkiger:
### 📣 Summary

URL pre-filled default values now persist across multiple submissions in offline-capable forms.

### 📖 Description

When a form URL includes pre-filled field values via query parameters (e.g. `?d[participant_id]=12345`), those values are now correctly re-applied each time the form resets after a successful submission. Previously, only the first submission in a session would carry the pre-filled value; subsequent submissions showed a blank field, requiring the user to manually refresh the page to restore the default.

### 💭 Notes

**Root cause:** After a successful submission, `_resetForm()` re-initialised the form from the blank XForm model, but never re-applied the URL defaults. The `settings.defaults` map was still in memory and the URL was unchanged, but nothing called `_prepareInstance` again.

**What changed:**

- `enketo-webform.js`: The URL-defaults logic is now split into two functions with distinct responsibilities:
  - `getValidatedDefaults(modelStr, defaults)` *(new, exported)* — the security gate. Validates each `d[path]=value` URL param against the form model using the existing `_isAllowedDefault` checks (blocks attributes, group nodes, the instance root, and protected ODK metadata fields like `instanceID`). Returns only the safe subset.
  - `_prepareInstance(modelStr, validatedDefaults)` *(now private)* — unchanged in behaviour, but now accepts pre-validated defaults rather than raw URL params, since the validation responsibility has been lifted out.
  - On form initialisation, both are called; `validatedDefaults` is passed into `controller.init()` alongside `instanceStr`.

- `controller-webform.js`: `_resetForm()` now re-applies `formData.validatedDefaults` after `form.init()` using `setVal` + `setAllVals()`. The controller itself has no knowledge of validation rules — it just applies what it was given.

- `url-defaults.spec.js`: Tests now cover `getValidatedDefaults` (where the security behaviour lives) rather than `_prepareInstance`.

### 👀 Preview steps

1. ℹ️ Have a form with a text field (e.g. `participant_id`) and a second unrelated text field
2. Obtain its Enketo URL and append `?d[participant_id]=TEST123`
3. Navigate to the online-offline form URL 
4. 🔴 **[on main]** Notice `TEST123` is pre-filled in `participant_id` on first load ✓
5. Fill in the second field and submit successfully
6. 🔴 **[on main]** Notice `participant_id` is now blank on the reset form — the default did not return
7. 🟢 **[on PR]** Notice `TEST123` is pre-filled again after submission, without needing to refresh
8. 🟢 **[on PR]** Submit again and confirm the second submission succeeds

### DIFF
--- a/packages/enketo-express/public/js/src/enketo-webform.js
+++ b/packages/enketo-express/public/js/src/enketo-webform.js
@@ -279,10 +279,11 @@ export function getValidatedDefaults(modelStr, defaults) {
     const validated = {};
 
     for (const path in defaults) {
-        if (!Object.prototype.hasOwnProperty.call(defaults, path)) continue;
-        const targetNode = model.node(path).getElement();
-        if (_isAllowedDefault(path, targetNode, model.rootElement)) {
-            validated[path] = defaults[path];
+        if (Object.prototype.hasOwnProperty.call(defaults, path)) {
+            const targetNode = model.node(path).getElement();
+            if (_isAllowedDefault(path, targetNode, model.rootElement)) {
+                validated[path] = defaults[path];
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


pauloamorimbr (approved):
LGTM! 
Tested and works great.

PR #1603, opened 2026-08-28
feat(transifex): automate transifex pull and push workflows DEV-2740

pauloamorimbr:
### 💭 Notes

This PR adds automation for Transifex:

- Push: whenever locales/src/en/translation.json (source strings) changes on main, pushes it to Transifex so translators see new/changed keys immediately.
- Pull: on a weekly schedule, pulls completed translations from Transifex and opens a PR with any updated locale files.
- Both workflows can also be triggered manually.
- Zulip notifications: a failure message on either workflow failing, and a success message on pull when a translations PR is actually opened (silent if there's nothing new).

⚠️  This cannot be tested as a PR, so we need to first merge this and then test it.


.github/workflows/transifex-pull.yml
```diff
@@ -0,0 +1,75 @@
...
+            pull-requests: write
+        outputs:
+            pr-operation: ${{ steps.cpr.outputs.pull-request-operation }}
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
+
+            - name: Install Transifex CLI
+              run: |
+                  curl -o- https://raw.githubusercontent.com/transifex/cli/master/install.sh | bash
```
jamesrkiger: Did you see they have a cli-action? Any reason not to use that?
https://github.com/transifex/cli-action

If there's an issue with the CLI action, though, I think it might good to grab a pinned release version rather than running the install script like this.

tx
jamesrkiger, line 1: We don't need this, do we?

jamesrkiger (approved):
Left a couple comments

.github/workflows/transifex-pull.yml
Akuukis: I recently automated transifex on KPI, GHA worked fine for me: see [push](https://github.com/kobotoolbox/kpi/blob/main/.github/workflows/locale.yml#L97-L102) and [pull](https://github.com/kobotoolbox/kpi/blob/main/.github/workflows/release-3-tag.yml#L159-L164) jobs.

.github/workflows/transifex-pull.yml
```diff
@@ -0,0 +1,75 @@
...
+# Requires:
+# - a `packages/enketo-express/.tx/config` file (generate locally with
+#   `tx add remote` once, then commit it - see tutorials/40-translate.md)
+# - a `TRANSIFEX_TOKEN` repository secret with a Transifex API token that has read
+#   access to the enketo-express project
+
+on:
+    schedule:
+        # Every Monday at 06:00 UTC
+        - cron: '0 6 * * 1'
```
Akuukis: I recommend using odd minutes, because GHA gets congested on exact hours and may timeout or just act weird.

.github/workflows/transifex-pull.yml
```diff
@@ -0,0 +1,75 @@
...
+                  echo "$GITHUB_WORKSPACE" >> "$GITHUB_PATH"
+
+            - name: Pull translations
+              working-directory: packages/enketo-express
+              env:
+                  TX_TOKEN: ${{ secrets.TRANSIFEX_TOKEN }}
+              # Only pull strings that have actually been translated, so untranslated
+              # keys keep falling back to English instead of being overwritten with
+              # empty/placeholder values.
+              run: tx pull --all --mode onlytranslated
```
Akuukis: KPI automation uses `--mode reviewed`, why not do the same for consistency?

.github/workflows/transifex-push.yml
```diff
@@ -0,0 +1,47 @@
...
+            - 'packages/enketo-express/locales/src/en/translation.json'
+            - 'packages/enketo-express/.tx/config'
+    workflow_dispatch: {}
+
+jobs:
+    push-source-strings:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
```
Akuukis:
```suggestion
              uses: actions/checkout@v7
```

.github/workflows/transifex-pull.yml
```diff
@@ -0,0 +1,75 @@
...
+    pull-translations:
+        runs-on: ubuntu-latest
+        permissions:
+            contents: write
+            pull-requests: write
+        outputs:
+            pr-operation: ${{ steps.cpr.outputs.pull-request-operation }}
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
```
Akuukis:
```suggestion
              uses: actions/checkout@v7
```

.github/workflows/transifex-push.yml
```diff
@@ -0,0 +1,47 @@
+name: Transifex - push source strings
+
+# Pushes the English source strings to Transifex whenever they change on main,
+# so translators immediately see new/changed keys to translate.
+#
+# Requires:
+# - a `packages/enketo-express/.tx/config` file (generate locally with
+#   `tx add remote` once, then commit it - see tutorials/40-translate.md)
+# - a `TRANSIFEX_TOKEN` repository secret with a Transifex API token that has push
+#   access to the enketo-express project
```
Akuukis:
```suggestion
```
unnecessary LLM self-commentary

.github/workflows/transifex-pull.yml
```diff
@@ -0,0 +1,75 @@
+name: Transifex - pull translations
+
+# Periodically pulls completed translations from Transifex and opens a PR,
+# so translation work done by the community actually lands in the repo.
+#
+# Requires:
+# - a `packages/enketo-express/.tx/config` file (generate locally with
+#   `tx add remote` once, then commit it - see tutorials/40-translate.md)
+# - a `TRANSIFEX_TOKEN` repository secret with a Transifex API token that has read
+#   access to the enketo-express project
```
Akuukis:
```suggestion
```
unnecessary LLM self-commentary

.github/workflows/transifex-push.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+    push:
+        branches:
+            - main
+        paths:
+            - 'packages/enketo-express/locales/src/en/translation.json'
+            - 'packages/enketo-express/.tx/config'
```
Akuukis: Should also add a test in the `install-build-test.yml` that the source strings are up to date, otherwise they will get stale and this won't trigger.

jamesrkiger (changes requested):
Could you also take a look at/address Kalvis' comments? They sound like good ideas to me.

jamesrkiger (approved):
LGTM

.github/workflows/transifex-push.yml
pauloamorimbr: This would not apply for Enketo, since the translation file is always directly updated by the developer, since in code we only use the translation keys.

PR #1607, opened 2026-09-08
fix(transifex): fix pull automation failures DEV-2740

pauloamorimbr:
### 📣 Summary

Fixes the Transifex pull automation added in #1603, which failed on its first scheduled/manual run, plus adds a CI guard to prevent the underlying issue from recurring.

### 💭 Notes

While debugging the failed [Transifex - pull translations](https://github.com/enketo/enketo/actions/runs/34258069615) run, found two issues:

1. **Auth failure on "Open pull request" step** (`fatal: unable to access ... 400`, `Duplicate header: "Authorization"`, exit 128).
   `actions/checkout` persists the job's `GITHUB_TOKEN` as a git credential (`http.extraheader`) by default. `peter-evans/create-pull-request` then configures its own credential the same way, and git ends up sending two `Authorization` headers, which GitHub rejects with a 400.
   Fix: set `persist-credentials: false` on the checkout step, so only `create-pull-request` manages git credentials.

2. **`tx pull` skipping files that should have been updated** (`Local file is newer than remote, skipping`).
   By default, `tx pull` compares the filesystem modification time of local files against Transifex's last-review date to decide whether to overwrite them. On a fresh `actions/checkout`, every existing file's mtime is reset to "now" (the checkout instant), which is always more recent than Transifex's last-review date — so every already-existing locale file was being skipped, regardless of actual translation staleness.
   Fix: add `--use-git-timestamps` to the `tx pull` args, so freshness is judged by the file's last git commit date instead of checkout mtime.

3. ⚠️  Important -> **New `check-translations` CI job** ([ci.yml](https://github.com/enketo/enketo/blob/main/.github/workflows/ci.yml)) fails a PR if any `translation.json` other than `locales/src/en/translation.json` is changed, since non-English translations should only ever be updated by the Transifex pull automation. The automation's own PR (branch `chore/transifex-translations`) is exempted from this check.

### 👀 Preview steps

1. Merge this PR
2. Manually trigger the [Transifex - pull translations](https://github.com/enketo/enketo/actions/workflows/transifex-pull.yml) workflow again and confirm the "Open pull request" step succeeds
3. Open a PR that hand-edits a non-English translation file and confirm the new `check-translations` job fails it


jamesrkiger (changes requested):
I want to give some consideration to the overall strategy here. We are telling people they shouldn't update transifex translations manually, but our config has `replace_edited_strings = false`, which means if someone does update them manually we are explicitly  going to try to keep those edits over whatever is in transifex. And then transifex CLI is checking timestamps to ensure it doesn't overwrite any of the edits we are trying to prevent people from making. So it feels like the configuration is pulling in opposing directions. I would think we should pick one: Either we anticipate making manual edits and ensure that transifex doesn't interfere with them, or we explicitly don't allow manual edits and therefore allow transifex to overwrite the translation files (obviously in the context of a PR). Note: In the latter case, I'd still be inclined to say we should drop the check-translations job. If the files are going to be overwritten by transifex, we don't need CI to protect them.

pauloamorimbr:
> I want to give some consideration to the overall strategy here. We are telling people they shouldn't update transifex translations manually, but our config has `replace_edited_strings = false`, which means if someone does update them manually we are explicitly going to try to keep those edits over whatever is in transifex. And then transifex CLI is checking timestamps to ensure it doesn't overwrite any of the edits we are trying to prevent people from making. So it feels like the configuration is pulling in opposing directions. I would think we should pick one: Either we anticipate making manual edits and ensure that transifex doesn't interfere with them, or we explicitly don't allow manual edits and therefore allow transifex to overwrite the translation files (obviously in the context of a PR). Note: In the latter case, I'd still be inclined to say we should drop the check-translations job. If the files are going to be overwritten by transifex, we don't need CI to protect them.

After talking during a meeting, I'm applying the changes we've decided.

jamesrkiger (approved):
LGTM

PR #1608, opened 2026-09-09
fix(xpath): encode digest() message as UTF-8 DEV-2814

### LINKED ISSUE #1605: digest() can produce incorrect SHA-256 values during Enketo editing
`digest()` can produce incorrect SHA-256 values during Enketo editing

## Describe the bug

When an XLSForm calculation calls `digest()` on a node reference, Enketo can calculate an incorrect SHA-256 value rather than the SHA-256 of the referenced field value.

There appear to be at least two related behaviours.

### 1. Direct node references can collapse to a constant incorrect digest

For example:

```text
digest(${some_field}, 'SHA-256', 'hex')
```

In our ODK Central workflow, opening an existing submission for editing through Enketo and saving it caused multiple unrelated calculated SHA-256 fields to be replaced by the same value:

```text
5d668fb1422a7dc905e48cf5addcd27069586fdbffd91c56951b2cbf2d0869f6
```

This occurred even though the underlying source fields contained different values and remained unchanged.

A previous ODK Forum report describes similar behaviour and suggested wrapping the node reference in `concat()`:

```text
digest(concat(${some_field}), 'SHA-256', 'hex')
```

### 2. The `concat()` workaround does not reliably hash Unicode strings

We subsequently constructed a controlled diagnostic XLSForm to test the workaround across ASCII strings, Unicode strings, string lengths up to 3000 UTF-8 bytes, and JSON-like values.

The results show that:

- `digest(${field}, ...)` fails after an Enketo edit for all tested values and produces the same incorrect constant digest;
- `digest(concat(${field}), ...)` works correctly for ASCII values, including strings of at least 3000 bytes;
- `digest(concat(${field}), ...)` still produces an incorrect SHA-256 for values containing non-ASCII Unicode characters;
- this Unicode failure occurs even for very short strings and is therefore not a string-length limitation.

For example, the test string:

```text
abc © — xyz
```

is 14 UTF-8 bytes.

Its correct SHA-256 is:

```text
03f921e8d84b143063d4a8cf2adfb944a8a50c909b04d1b513b9eed20ad499fb
```

ODK Collect calculated this value correctly.

After opening the same submission in Enketo, changing an unrelated edit-marker field, and saving the record, the `digest(concat(...))` result no longer matched the SHA-256 of the unchanged retained source value.

## Environment / context

- ODK Central
- Existing submissions opened and edited through Central's web editing interface / Enketo
- SHA-256 output requested as hexadecimal
- September 2026
- Original records collected using ODK Collect
- Retained historical submission versions independently audited using Python SHA-256 calculations
- Source strings were extracted directly from the retained historical XML for independent verification

## Minimal reproduction

A diagnostic XLSForm contains fixed source strings together with both digest variants:

```text
digest(${input_text}, 'SHA-256', 'hex')
```

and:

```text
digest(concat(${input_text}), 'SHA-256', 'hex')
```

Each test also contains an independently precomputed expected SHA-256.

The matrix includes:

- short ASCII;
- ASCII strings of 1024, 2048, 2500, 2671 and 3000 bytes;
- short Unicode;
- Unicode strings around 2671 bytes;
- JSON-like ASCII strings;
- JSON-like Unicode strings at several sizes.

## Suggested reproduction

1. Publish the diagnostic XLSForm.
2. Submit it using ODK Collect.
3. Confirm the calculated digests.
4. Open the submitted record for editing through ODK Central / Enketo.
5. Change only the dedicated edit-marker field.
6. Save the edit.
7. Compare the calculated SHA-256 fields with independently calculated SHA-256 values for the actual retained source strings.
8. Repeat if desired with another change to the edit-marker field.

### Direct node reference

After an Enketo edit, every tested expression of the form:

```text
digest(${field}, 'SHA-256', 'hex')
```

failed.

Across unrelated source values, the resulting SHA-256 collapsed to:

```text
5d668fb1422a7dc905e48cf5addcd27069586fdbffd91c56951b2cbf2d0869f6
```

### `concat()` workaround

Expressions of the form:

```text
digest(concat(${field}), 'SHA-256', 'hex')
```

behaved as follows:

```text
Short ASCII                 PASS
ASCII 1024 bytes            PASS
ASCII 2048 bytes            PASS
ASCII 2500 bytes            PASS
ASCII 2671 bytes            PASS
ASCII 3000 bytes            PASS

Short Unicode               FAIL
Unicode ~2671 bytes         FAIL

ASCII JSON-like text        PASS
Unicode JSON-like text      FAIL
```

This indicates that string length itself is not the limiting factor.

The distinguishing characteristic of the remaining failures is the presence of non-ASCII Unicode characters.

Importantly, the actual source strings retained in Central did not change between the original submission and the Enketo-edited version.

## Behaviour observed in a larger real form

We originally detected the issue because the form contains cryptographic provenance checks.

The original submission independently verified correctly:

```text
Stored payload SHA:
557f5069cc8bef77eade4b7a92b78500f9c900576e059ca308cf4dc3e7105b0d

Independently reconstructed SHA:
557f5069cc8bef77eade4b7a92b78500f9c900576e059ca308cf4dc3e7105b0d

Result:
INDEPENDENTLY_VERIFIED_ATTESTED_VERSION
```

After editing the submission in Central/Enketo, the form recalculated its stored payload hash as:

```text
5d668fb1422a7dc905e48cf5addcd27069586fdbffd91c56951b2cbf2d0869f6
```

Independent reconstruction from the retained source values instead produced:

```text
312104bdd3d7aac2aa6171067c5e0a18779d7095dd0ee713c452db0215ca9f12
```

Seven independent text/JSON digest fields initially mismatched:

```text
auth_json_sha256
photo_json_sha256
location_json_sha256
polygon_json_sha256
conversation_transcript_sha256
conversation_json_sha256
fourat_json_sha256
```

After changing those calculations to use:

```text
digest(concat(${field}), 'SHA-256', 'hex')
```

six of the seven mismatches disappeared.

The remaining mismatch was:

```text
fourat_json_sha256
```

The underlying 4AT JSON contained Unicode characters including an em dash (`—`) and copyright symbol (`©`).

The source JSON remained unchanged between historical Central versions, but Enketo recalculated its SHA-256 incorrectly.

A separately generated SHA-256 of an image attachment continued to match the actual attachment bytes throughout, indicating that the source record itself had not generally been corrupted.

## Expected behaviour

For:

```text
digest(${some_field}, 'SHA-256', 'hex')
```

Enketo should calculate the SHA-256 of the current string value represented by `${some_field}`.

Likewise:

```text
digest(concat(${some_field}), 'SHA-256', 'hex')
```

should calculate the SHA-256 of the resulting string according to the documented character encoding.

For a given string value, the digest should agree between:

- ODK Collect;
- Enketo;
- an independent standards-compliant SHA-256 implementation.

Unicode strings should be hashed consistently, with their encoding behaviour clearly defined.

## Actual behaviour

During Enketo editing:

1. direct-node `digest()` expressions can produce an incorrect constant SHA-256 unrelated to the underlying source value; and
2. wrapping the node reference in `concat()` corrects the problem for tested ASCII strings but still produces incorrect digests for tested Unicode strings.

The calculation succeeds syntactically, so the user receives no indication that the hash is incorrect.

## Impact

This may affect forms that use `digest()` for:

- checksums;
- pseudonymous identifiers;
- data-integrity checks;
- authentication workflows;
- cryptographic provenance;
- deduplication;
- record linkage;
- audit trails;
- identifiers derived from form content.

The risk is particularly significant during editing because a user may change an unrelated field and Enketo can silently replace previously correct calculated digest values.

The resulting record can therefore contain checksum fields that no longer correspond to the data from which they purport to have been calculated.

## Current workaround / mitigation

Wrapping the argument in `concat()`:

```text
digest(concat(${some_field}), 'SHA-256', 'hex')
```

appears to avoid the direct-node problem for ASCII strings.

However, our diagnostic tests show that this is not a complete workaround because Unicode-containing values can still produce incorrect hashes.

For our provenance system we are therefore not treating Enketo-calculated hashes on edited records as authoritative.

Instead, an independent verifier:

1. retrieves the retained historical submission;
2. retrieves a signed machine-readable commitment recipe;
3. reads the actual source field values and attachment bytes;
4. independently recomputes all component hashes;
5. reconstructs the canonical committed payload;
6. independently calculates its SHA-256.

This avoids depending on Enketo's recalculated digest fields for integrity verification.

## Available supporting material

We can provide:

1. the minimal diagnostic XLSForm containing the ASCII / Unicode / size test matrix;
2. original and Enketo-edited historical submission XML;
3. the independently calculated Python SHA-256 values;
4. the larger provenance audit logs showing the discrepancy before and after Enketo edits.

jnm:
### 📣 Summary

Fixes the `digest()` function, which calculated the wrong hash for any value containing non-ASCII characters.

### 📖 Description

Any digest of a non-ASCII value calculated before this fix is wrong and will change once this ships. That is the intended correction, but it does mean previously stored digests of non-ASCII values won't compare equal to newly calculated ones.

`digest()` calls node-forge's `md.update(message)` without the second `encoding` argument. node-forge represents binary data as a "byte string" — a JS string in which every code unit must be ≤ `0xFF` and stands for one byte — and only performs UTF-8 encoding when explicitly passed `'utf8'`. Without it, the message was consumed as though it were already a byte string, so each UTF-16 code unit was read as one byte. `md.messageLength` likewise counted code units rather than bytes, so the padded length field described a shorter message than the real one.

That produced three distinct corruptions:

1. **`0x80`–`0xFF`**: the code unit fits in a byte slot, yielding Latin-1 rather than UTF-8. `é` hashed as `E9`, not `C3 A9`.
2. **Above `0xFF`, at word offset 0**: `getInt32()` packs four code units per 32-bit word as `charCodeAt(i) << 24 ^ …`, so `code << 24` overflows and the high byte is discarded. `—` (`U+2014`) hashed as `14`.
3. **Above `0xFF`, at word offsets 1–3**: the smaller shift puts the high byte into a *neighbouring* slot, where the `^` XORs it into whatever is already there — so an adjacent character's byte changes. In `a—`, the em dash flipped the `a`: `0x61 ^ 0x20 = 0x41`.

Worked example, `abc © — xyz` (verified against a reimplementation of forge's packing, which reproduces its output exactly):

```
correct (UTF-8):  61 62 63 20 C2 A9 20 E2 80 94 20 78 79 7A   (14 bytes)
what was hashed:  61 62 63 20 A9 00 14 20 78 79 7A            (11 bytes)
```

In word 1 the em dash's leading `0x20` XOR-cancelled the space preceding it, so both characters vanished:

```
©(0xA9) << 24  =  A9 00 00 00
 (0x20) << 16  =  00 20 00 00
—(0x2014) << 8 =  00 20 14 00
 (0x20)        =  00 00 00 20
                  ───────────
XOR            =  A9 00 14 20
```

The fix passes `'utf8'`, so forge runs `util.encodeUtf8()` first and hashes the correct bytes. Since ASCII is one code unit *and* one UTF-8 byte per character with no high bits to overflow or collide, ASCII output is byte-identical before and after — the existing assertions don't move.

Two incidental cleanups in the same function: the hex branch called `md.digest()` a second time rather than reusing the already-finalized buffer (harmless, as forge's `digest()` clones state, but it redid the finalize work), and the `!encoding` check was dead code because `encoding` is defaulted to `'base64'` on assignment.

### 💭 Notes

New test coverage: md5 / sha-1 / sha-256 / sha-384 / sha-512 in hex for a mixed 2- and 3-byte string, base64 both explicitly and by default, a wholly non-ASCII Cyrillic string, a surrogate pair (`🎉`, which must encode to four bytes), and a non-ASCII value reached through a dereferenced XPath. All ten new assertions fail on `main` and pass here; the six pre-existing ones are unchanged.

** :robot: This PR (code and description) was written over a 2.5h session with Claude Opus 5 :robot: **

### 👀 Preview steps

Minimal XLSForm — one `calculate` and one `note`, no data entry needed:

| type | name | label | calculation |
| --- | --- | --- | --- |
| calculate | sha | | `digest('abc © — xyz', 'SHA-256', 'hex')` |
| note | out | `SHA-256: ${sha}` | |

1. ℹ️ On your XLSForm-compatible platform of choice (e.g. KoboToolbox, Ona, ODK Central), upload the XLSForm above
2. Use your platform's preview function to open the form in Enketo
3. 🔴 [on main] notice the digest is `6f5a97d90aff1f0d0a4216d273b7e4b14435c1d25c5e0f2b80c5922b27412e7b`
4. 🟢 [on PR] notice the digest is `03f921e8d84b143063d4a8cf2adfb944a8a50c909b04d1b513b9eed20ad499fb`, which is the SHA-256 that ODK Collect, `python3 -c "import hashlib; print(hashlib.sha256('abc © — xyz'.encode()).hexdigest())"`, and `sha256sum` all produce
5. 🟢 [on PR] confirm ASCII is unchanged: `digest('abc', 'SHA-256', 'hex')` is `ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad` both before and after


pauloamorimbr (approved):
LGTM! 👍🏻

